PR #1254, opened 2014-04-06
Dev roelski newanidbupdate

Roelski:
Change behaviour of update_anidb: first fill db, then update records. That way always filled db


Roelski:
Ok, thanx.
I use a mix of notepad++, vi and winscp editor as I'm working from a 
linux based nas.
Not sure how I can see in those editors what I did wrong, but I'll check 
on it.

b3rs3rk schreef op 13-4-2014 17:51:

PR #1583, opened 2015-01-23
Update populate_tvrage.php to be a bit more informative

DariusIII:


niel:
Can you redo this as a new PR without the doubled formatting changes. It makes it too much hassle to check it when 99% of the changes are formatting and it's a job of making sure the end result is the same as it started.
Also I'm really uncomfortable with all those "Merge remote-tracking branch 'origin/dev' into dev" commits. They really shouldn't be included in a PR, it implies that there is some change hiding in them that is different from the branch the PR is against and that means a lot of checking to verify it.

niel:
If you're having trouble getting rid of the "Merge remote-tracking branch 'origin/dev' into dev" get hold of me in dev's IRC and we can work that out together.

DariusIII:
Yes, i would love to get rid of those.

PR #1600, opened 2015-02-21
Add support for in DB collections (grouping) regex.

kevinlekiller:
Adds support for collections regex (grouping headers together to form releases) in the DB, while still supporting the "hard coded" regex.

These regex are not for naming or renaming releases.

Regex added to the DB will be used first, falling back on the "hard coded" regex if no matches are found.

See the hints here for more info on the regex : https://github.com/kevinlekiller/nZEDb/blob/patch2/www/themes_shared/templates/admin/collectionsregex-edit.tpl#L23

Newly added regex need to be added to the tsv file so new installs can get access to them, and they need to be in a patch, so current installs get access to them, or you can "hard code" them, but if you put them in the DB they can be edited by the users easily.


Bart39:
testing

PR #1632, opened 2015-03-29
Add X-DNZB-Failure header support.

DariusIII:


Bart39:
branched as dev-xdnzb for testing

Bart39:
updated dev-xdnzb as per updated PR

PR #1650, opened 2015-05-07
Adding ReCaptcha support

zetas:
If site-admin adds ReCaptcha site and secret keys in the admin->site_edit page, captcha's will display on every page that accepts unauthenticated user input (login, register, contact, forgottenpassword). If one or both of the keys are empty nothing is displayed, no errors. Seamless "failure". 

This PR includes the ReCaptcha PHP library by Google as well as a wrapper class, Captcha() that integrates with the site. There were many ways to accomplish this but I tried to stick to DRY as much as possible which is why I opted to process everything in BasePage instead of copying and pasting the same code in each of the 4 pages. It adds some slight overhead to every page load but only inasmuch as instantiating an empty class. 

You'll see some type checking going on (which I know we use sparingly) but I think it's justified. As you can see, $captcha->getError() outputs bool false if no errors. If something should happen and a blank error comes through (like if they update their error messages and we don't support it) It should still display an error and not allow the page transaction through. If you guys think otherwise, I can change it.

I tried to implement this in a way that affected as few pages as possible. It still requires a line added to each page and template but I didn't see a way to do it otherwise for the same amount of effort.

Let me know if anything needs to be added or changed.

Also, FWIW I did test with every single template and they all work as expected both with and without the recaptcha enabled.

### DIFF
--- a/libs/ReCaptcha/ReCaptcha.php
+++ b/libs/ReCaptcha/ReCaptcha.php
@@ -1,0 +1,97 @@
+<?php
+/**
+ * This is a PHP library that handles calling reCAPTCHA.
+ *
+ * @copyright Copyright (c) 2015, Google Inc.
+ * @link      http://www.google.com/recaptcha
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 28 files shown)


kevinlekiller:
Tested on my end, didn't find any issues.

Bart39:
i shall merge then :)

PR #1654, opened 2015-05-10
Fix Popporn error - causing xxx not to post process

taskym:
This will fix an non-object fatal error caused during post processing XXX.


kevinlekiller:
#1655

PR #1656, opened 2015-05-10
Move files around in misc/testing folder.

kevinlekiller:
Delete scripts that were not useful any more.

Move existing scripts into proper folders.

misc/testing/IRCScraper/ moved to misc/IRCScraper/ -> It will try to use the settings file from misc/IRCScraper/ and fall back on misc/testing/IRCScraper/ if not found.


kevinlekiller:
Note, the patches recommended by scrutinizer are not useful.

PR #1658, opened 2015-05-11
Correct logout cookie deletion and cookie setting vars.

Enverex:
Cookies won't be deleted properly on sub-folders/HTTPS setups as the deletion code doesn't follow the same rules as the creation code for the cookies. This fixes that. Second block of changes correct null and bool variable usage.


nzedb/Users.php
```diff
@@ -838,8 +838,9 @@ public function logout()
 	{
 		session_unset();
 		session_destroy();
-		setcookie('uid', '', (time() - 2592000));
-		setcookie('idh', '', (time() - 2592000));
+                $secure_cookie = (isset($_SERVER['HTTPS']) && $_SERVER['HTTPS'] == 'on' ? '1' : '0');
```
kevinlekiller: Indentation issues (spaces).

PR #1660, opened 2015-05-11
Fix reported issues in Forking and ProcessReleases.

DariusIII:


testarea1:
After I pulled the Captcha move under 2bad811e3cd7f57333b449997211b03e46a39f01, I added
```
use \nzedb\Captcha;
```
in www/pages/BasePage.php and was able to bring up the site again. Was this updated class path part of a new PR anywhere? I only see the one from earlier with the old "\nzedb\controllers\Captcha" path before move.

niel:
@testarea1 Added the use in the last batch of fixes.

DariusIII:
As niel fixed the issues, this PR is irrelevant now.

PR #1665, opened 2015-05-11
Fixing typo to allow for self-signed SSL certs with SabNZBd.

zetas:
Verify peer was a typo and was breaking the ability for users to use self-signed cert SSL with sab (which is the default sab ssl setup)


Bart39:
closing as this has already been merged

PR #1672, opened 2015-05-14
Implement local temporary caching for Amazon lookup failures.

Enverex:
Temporarily store the music/games/books we've looked up this run that failed so that if we query them again we can immediately return the failure rather than wasting time repeating ourselves.


Enverex:
This appears to be a duplicate.

PR #1680, opened 2015-05-16
Broken install fix. iconv check. Use iconv for converting CP437.

kevinlekiller:
Installation pages were broken due to namespace changes.

Add check for iconv on install, this is required in our INSTALL document: https://github.com/nZEDb/nZEDb/blob/master/INSTALL.txt#L25

Use iconv instead of PHP for converting CP437 to UTF-8, 100-1000x faster on various NFO's I have tested. For example, an nfo with 40k characters would take 0.2 seconds with PHP, with iconv it takes 0.0003 seconds.


www/install/templates/step1.tpl
```diff
@@ -1,249 +1,251 @@
...
-				{$fixString}
-				</span>
+	<tr>
+		<th>check</th>
+		<th style="width:75px;">status</th>
+	</tr>
+	<tr class="alt">
+		<td>Checking for iconv support:{if !$cfg->iconvCheck}<br /><span class="error">The PHP installation lacks support for iconv.</span>{/if}</td>
+		<td>{if $cfg->iconvCheck}<span class="success">OK</span>{else}<span class="error">Error</span>{/if}</td>
+	</tr>
```
kevinlekiller: Line 5 to 9 are the only real changes, the rest was an indentation fix.

Indentation fix: https://github.com/kevinlekiller/nZEDb/commit/694cc37e172be5d06e5270f53fde4f5969c3905f

Line 5 to 9 addition: https://github.com/kevinlekiller/nZEDb/commit/fe4017932225871f1115dd017376617756c9d644#diff-8dc357d49a7587abf80090cc2bc9c390L5

kevinlekiller:
Don't pin that shit on me scrutinizer, lick my balls.

PR #1681, opened 2015-05-17
IRCScraper fixes.

kevinlekiller:
Fix include path from changing the folder location.

Fix namespace issue from changing the namespace.


Wally73:
that fixed it thx ;)

Issue #1684, opened 2015-05-17
Renamed Releases Lose Leading "S" Character


Enverex:
I've noticed this a few times and it seems oddly specific so I can only assume it's a bug. What is happening is that some releases are being renamed correctly, but they lose the leading S on the Author or Title. So for example "Bob Bobson - Summer Holiday" would become "Bob Bobson - ummer Holiday".

An example I just saw was:
![image](https://cloud.githubusercontent.com/assets/958670/7670676/33088db2-fca6-11e4-8454-e0c196adb0f1.png)

Where Steve Vai became "teve Vai". The renamer in question here was "sorter audioNFO".

Also to clarify, I've only ever seen this happen with the S character, no other letters seem to be affected.

Bart39:
merged the fix

PR #1685, opened 2015-05-17
Update the audioNFO regex to be slightly less rubbish

Enverex:
It should now match more often and not lose characters from the start of the artist or the album (specifically K and S).


kevinlekiller:
#1684

PR #1699, opened 2015-05-21
Remove superfluous searchname param from href links.

kaibosh:
Prevents searchname from appearing in browser history and proxy logs.

For now only patched the default theme. If this is the correct way to implement this change, I'll do another PR for the other themes.


Enverex:
As far as I'm aware this is deliberate. It's part of the SEO optimisation for the indexer (which seems a bit weird in the first place). This should likely be made optional rather than outright removed though.

kaibosh:
Hadn't considered that use case. Agreed, removing this entirely is a little heavy-handed. Could be considered as a future nice to have option, as maintaining as local mod is quite easy.

PR #1715, opened 2015-06-06
fixed #1713

### LINKED ISSUE #1713: Bug - Select all + delete in adv search not working
When using the adv search and clicking the select all checkbox and then clicking the admin delete button it only refreshes the page, all results are still there. 

However I think I found the issue.

When clicking the select all checkbox it also checks two checkboxes in the search box, after unchecking both of them the delete button is now functional.

bug

It may be related to #283

rainje:
changed jquery selector for alpha & default theme

### DIFF
--- a/www/themes/Default/scripts/utils.js
+++ b/www/themes/Default/scripts/utils.js
@@ -38,7 +38,7 @@ jQuery(function ($) {
     });
 
     $('.nzb_check_all').change(function () {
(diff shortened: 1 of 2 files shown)


niel:
We do not accept PRs against master, can you redo this PR based off the dev branch please.

Bart39:
manually applied fix in dev

PR #1726, opened 2015-06-18
Support for downloading last rar or zip file when post processing.

kevinlekiller:
Some posters are now putting some of the files (like nfos) in the last rar / zip files, so we miss out on getting those.


nzedb/processing/post/ProcessAdditional.php
```diff
@@ -796,14 +801,13 @@ protected function _getNZBContents()
 		}
 
 		// Get a list of files in the nzb.
-		$this->_nzbContents = $this->_nzb->nzbFileList($nzbContents);
+		$this->_nzbContents = $this->_nzb->nzbFileList($nzbContents, [false, true]);
```
kevinlekiller: This should be [ 0=>false, 1=>true]

Although you wrote you would change from numeric keys to string keys.

PR #1727, opened 2015-06-19
Fix the nzbFileList function, as it was causing  PHP Fatal error:  Un…

DariusIII:
…supported operand types

### DIFF
--- a/nzedb/NZB.php
+++ b/nzedb/NZB.php
@@ -334,7 +334,7 @@ class NZB
 	 *
 	 * @access public
 	 */
-	public function nzbFileList($nzb, array $options = null)
+	public function nzbFileList($nzb, array $options = [])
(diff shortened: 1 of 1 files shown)


Wally73:
i'm running the latest dev for the last 3 hours and  i haven't gotten any php fatal errors (or any error) 
PHP 5.6.9-0+deb8u1 (cli) (built: Jun  5 2015 11:03:27)

DariusIII:
Have you turned on the new rar option?

Wally73:
yes i have that enabled

http://gyazo.com/408785f376f10ecdbed339ca54938345

Wally73:
needs to be merged indeed. Error showed up in php_error.log and the list contents of the rar gave an empty page

PR #1734, opened 2015-06-21
Add option to admin area to allow/dissalow users to select their site…

DariusIII:
… theme.

### DIFF
--- a/resources/db/patches/mysql/+1~settings.sql
+++ b/resources/db/patches/mysql/+1~settings.sql
@@ -1,0 +1,1 @@
+INSERT INTO settings (section, subsection, name, value, hint, setting) VALUES ('site', 'main', 'userselstyle', '0', 'Allow users to change the site theme or not.', 'userselstyle');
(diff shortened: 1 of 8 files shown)


niel:
I do not see the point of this option.
On a personal level, I do not believe in removing options from users and forcing them to work they way the admin insists.
If you *must* remove the options, remove the templates so they're not available.

DariusIII:
This option is mainly pointed at admins that customize their templates. As per all the settings, nobody forces admins to use it. While i understand your point and agree to some level, it is still a viable option.

niel:
As I said, the option already exists. The admin can remove the templates if they want to prevent their use. Alternatively they could merge this PR themselves.
I don't believe this should exist in the core. It is for individual site admins to customise this kind of option themselves.

niel:
I would be more inclined to consider an option that restricts the templates individually. In effect virtually deleting them from the available ones.

Fossil01:
It's such a small change which might help a lot of site owners. Why not just merge it and be done with it? No effort at all. I for example like to keep the default/alpha theme around when I need to test something real quick.

PR #1747, opened 2015-07-02
API-Change regarding JSON-Output

sebst3r:
With this change we make sure, that "/api?t=caps&o=json" actually delivers an JSON-Output (before it was just XML, even if we used "&o=json"). For that reason the "caps.php" is new.


sebst3r:
fixed the duplicate code, renamed the class to "capabilities"

sebst3r:
added a description to the getForMenu method explaing what it is supposed to do + also fix the @param which doesn't exist in the sig

www/pages/api.php
```diff
@@ -79,18 +80,27 @@
 if ($uid != '') {
 	$page->users->updateApiAccessed($uid);
 	$apiRequests = $page->users->getApiRequests($uid);
-	if ($apiRequests > $maxRequests) {
-		showApiError(500, 'Request limit reached (' . $apiRequests . '/' . $maxRequests . ')');
+	if ($apiRequests['num'] > $maxRequests) {
```
kevinlekiller: This doesn't work, the method doesn't return a array. https://github.com/nZEDb/nZEDb/blob/dev/nzedb/Users.php#L1345

www/pages/api.php
```diff
@@ -79,18 +80,27 @@
...
 $releases = new Releases(['Settings' => $page->settings]);
 
-$page->smarty->assign('extended', (isset($_GET['extended']) && $_GET['extended'] == 1 ? '1' : '0'));
-$page->smarty->assign('del', (isset($_GET['del']) && $_GET['del'] == 1 ? '1' : '0'));
+if (isset($_GET['extended']) && $_GET['extended'] == 1) {
+	$page->smarty->assign('extended', '1');
+}
+if (isset($_GET['del']) && $_GET['del'] == 1) {
+	$page->smarty->assign('del', '1');
+}
```
kevinlekiller: These changes causes those variables (del/extended) to be unset.

www/themes/Default/templates/frontend/apicaps.tpl
```diff
@@ -1,22 +1,24 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <caps>
-	<server appversion="{$site->version}" version="0.1" title="{$site->title|escape}" strapline="{$site->strapline|escape}" email="{$site->email}" url="{$serverroot}" image="{$serverroot}themes_shared/images/logo.png" />
-	<limits max="100" default="100"/>
+    <server appversion="{$serverconf.version}" version="0.1" title="{$serverconf.title|escape}" strapline="{$serverconf.strapline|escape}" email="{$serverconf.email}" url="{$serverconf.url}" image="{$serverconf.image}" />
```
kevinlekiller: Bad indentation here.

www/pages/api.php
```diff
@@ -215,12 +225,36 @@ function($release) {
...
+
+			 //create array
+			 $retJson = array();
+			 //write capability information
+			 $retJson = $caps;
+			 //append category information
+			 $retJson["categories"] = $cats;
+			//use json_encode
+			 header('Content-type: application/json');
+			 echo json_encode($retJson);
```
kevinlekiller: Bad spacing, also many of these lines can be combined:

`header('Content-type: application/json');`
`echo json_encode(['categories' => (new Capabilities(['Settings' => $page->settings]))->getForMenu()]);`

www/pages/api.php
```diff
@@ -79,18 +80,27 @@
...
+}
 
 // Output is either json or xml.
-$outputXML = (isset($_GET['o']) && $_GET['o'] == 'json' ? false : true);
+$outputXML = true;
+if (isset($_GET['o'])) {
+	if ($_GET['o'] == 'json') {
+		$outputXML = false;
+	}
+}
```
kevinlekiller: This takes 6 lines instead 1 and does the same.

PR #1750, opened 2015-07-07
Show password on screen after reset

Fossil01:
Displays the password to the user after clicking the link in their e-mail. To display in template use something like:

```html
    {if $notice != ''}
            <div class="alert alert-info">{$notice}</div>
    {/if}
```


Fossil01:
Better?

niel:
I'm confused by the two commits. Did you do this in a branch, then merge into dev for the PR? If so that's not how we do things. Do the PR from your branch, so that merging it in can be done in isolation easier.

PR #1761, opened 2015-07-17
Dev dariusiii themefix

DariusIII:
I don't know if these niels merged commits will cause problems. If they wil, i will close this and on monday do another PR.


Bart39:
this breaks a load of stuff, can you redo without the extra commits

DariusIII:
I will redo this on moday from work.

PR #1764, opened 2015-07-18
Change max.headers.iteration setting, add admin editable value to adm…

DariusIII:
…in site edit.


DariusIII:
I have updated the PR as per requirements.

PR #1785, opened 2015-07-31
Allow max parts to be specified in TMUX setting.  Similar to existing…

bfincher:
… max collections setting

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -13,6 +13,8 @@ For details of when a commit makes it into the master branch see the ReleaseNote
 	* Mrg: PR 1784 from iguyking - Create a new Ubuntu 14.04.02 install instructions.
 	* Mrg: PR 1780 from darius - Add send to cart and send to my queue buttons to rest of the pages.
 	* Mrg: PR 1779 from darius - Change the search page and results display, add send to cart and send to my queue buttons to movies page.
+
+
 2015-07-25 niel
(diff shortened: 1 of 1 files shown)


Bart39:
not sure on this change (qqually i am not sure about the existing functionality) as stopping bin / backfill updates will just leave you with more incomplete collections.
will take a steer from the other devs

DariusIII:
Specifying max parts and stopping updates, in this context, is counter productive.

bfincher:
We already have the capability to stop updates based on max collections.  This is necessary when doing massive backfills and imports to keep the parts table from growing to tens of millions of rows and causing performance to grind to a halt.

Bart39:
yes i know, but as i said i don't like how that is implemented either, it's counter-productive as the only way you'll clear parts / binaries / collections is by creating complete collections which means you need to carry on downloading parts

bfincher:
I think you would be 100% correct if not for the issue of lots of collections never being considered "completed" until the 2 hour interval has passed.

Bart39:
thats a side issue and is mostly caused by subject missing the parts count (which we can't get around).
anyways as i said i'll discuss with the other dev's

bfincher:
Thanks.  And I understand if you want to reject this one based on the dislike of the original feature.

PR #1787, opened 2015-08-02
Changed GET to REQUEST

Fossil01:
It now allows both GET and POST requests to add stuff to cart.


Fossil01:
nvm, pushed wrong file. this is crap.

DariusIII:
Btw, REQUEST is less secure than separately using GET or POST.

PR #1789, opened 2015-08-03
Remove getQueue function from utils.js as it was interfering with que…

DariusIII:
…ue management in charisma and omicron templates.


niel:
I fail to see how the function can be the problem, are you sure that it is not just a symptom of another problem?
In other words, if the function isn't called, how can it be a problem? If it IS being called and you have removed it, then an error is happening somewhere else and might be hidden (but not fixed) by this.

DariusIII:
That function is not used at all, but interferes with the function of the same name in NZBget.php which IS used for queue management, and as getQueue id called in utils.js (https://github.com/nZEDb/nZEDb/blob/dev/www/themes/omicron/scripts/utils.js#L886) it overlaps with the function that sneeds to be used: https://github.com/nZEDb/nZEDb/blob/dev/nzedb/NZBGet.php#L412

The one in utils.js provided with omicron and charisma templates is actualy a leftover from newznab i missed to remove (and nntmux does not use it either)

niel:
Ahh, that makes sense now.

PR #1796, opened 2015-08-05
Dev

terbonium:
DB schema fixes. most of these are based on unnecessary use of UTF8 charset where only latin1 data is stored. this is a larger problem when these columns are indexed because UTF8 requires 3bytes per character in an index vs 1byte for a latin1.  Index are also fixed width and an index on a varchar(255) will always be 768 bytes wide regardless of the data stored.  Switching to latin1 reduces that to 255 bytes

*binaries.binaryhash is a 32 character (16bit) md5 string.  This column can be converted to binary(16).  I'll submit a fix for this after I review the php.  This would shrink the index size in half.  For large installs this should be another great savings.


niel:
Like the changes in general, but a fw changes needed.
From PRs we need sql patches to be numbered +1~<table>.sql, +2~<table>.sql, etc. We have a script that automatically converts them to the appropriate number when we merge them into dev.

PR #1800, opened 2015-08-06
remove redudant IRCScraper in misc/testing

bfincher:


niel:
First one is already committed. I'll manually add the settings file to .gitignore

PR #1808, opened 2015-08-10
store predb title hashes as varbinary(20) for quicker lookups.

terbonium:
no need to use a fulltext index

if needed hash varbinary(20) column can be expanded in the future to hold larger hash strings


kevinlekiller:
Merged. dafa15644a5c95d96fb30328ea54b020bf13bb55

scottia:
debug while applying /var/www/nZEDb/resources/db/patches/mysql/0369~predb.sql patch

Info: COMMENT: This should be a quick patch.
Info: COMMENT: You can verify them by running this command in MySQL: SHOW TRIGGERS LIKE 'predb' \g
Debug: DEBUG: Delimiter switched to $$
Debug: DEBUG: Delimiter switched to ;

Did it hurt anything or fail ??

kevinlekiller:
it's ok

PR #1843, opened 2015-08-24
Add: Sphinx Full-Text Search for Filenames

ruhllatio:
PR Adds full text sphinx search compatibility for found file names in release_files table.  Currently only works when selecting Advanced Search and populating the Filename field.


scottia:
Push IT !!  That is a very cool add by the way.  Thanks :)

ruhllatio:
Still working some bugs out.  Patience.

PR #1861, opened 2015-09-09
Blacklist activity date

bfincher:
Display the last date in which a binaryblacklist was used to black list or white list a part.  This could be helpful in determining which entries are actually being used and matching incoming parts.


nzedb/Binaries.php
```diff
@@ -1330,6 +1338,20 @@ protected function _retrieveBlackList($groupName)
...
 	/**
+	 * Update the last_activity date on the blacklist
+	 * @param int $blackListId The ID of the blacklist entry
+	 * @return void
+	 */
+	public function updateBlacklistActivity()
+	{
+		foreach ($this->_binaryblacklistIdsToUpdate as $bid) {
+			$this->_pdo->queryExec(sprintf('UPDATE binaryblacklist set last_activity = NOW() where id = %d', $bid));
+		}
```
kevinlekiller: Instead of looping, maybe concatenate all the id's in a string, it would make less queries.

` WHERE id IN (%s)', implode(',', $this->_binaryblacklistIdsToUpdate)`

bfincher:
I optimized per your suggestion.

niel:
Thanks for the PR.
I've merged with a couple minor modifications: replaced the updateBlacklistActivity method with inlined version, and renamed the new class field to match our code standards.

PR #1867, opened 2015-09-10
Add Regexes for Korean Newsgroups

izolight:
I created some regexes that help finding releases in korean Newsgroups.


izolight:
How come the collections are created without those?

DariusIII:
They are, that is true, but it would be nice from you to add korean groups specific regexes where they don't exist. I did not mean to piss you off or anything, it was just a badly worded suggestion.

izolight:
Oh I see. I guess my answer was equally badly worded, just trying to understand how nzedb works :). What kind of regexes are used if I don't have group specific collection regexes, so I don't create duplicates?

kevinlekiller:
Thanks izolight, can you possibly remake this without those extra commits from @bfincher and @DariusIII ? Would make it easier to see what is actually changed.

PR #1869, opened 2015-09-10
I forgot to add the new binary black list column in the schema

bfincher:


Bart39:
this will need redoing as there are conflicts

Wally73:
thought that Kevin already fixed something (but i can be mistaken)

Bart39:
he fixed the empty array issue which is why this pr will need redoing for the other fixes

Wally73:
ah ok my bad Bart39 ;) (and yeah that array fix helped me :D )

kevinlekiller:
Added.

PR #1885, opened 2015-09-24
Add statistical info for grabbed/failed releases to cover, browse and search pages.

DariusIII:


Bart39:
is the new column needed ? couldn't you achieve the same by adding a function to Releases class that returns the failed count from dnzb_failures by guid directly instead

DariusIII:
That i how i started, but realized that it is more effective, on the longer run, to have a separate column in releases which will have the info, than to do a count on dnzb_failures table every time info is needed..

DariusIII:
I have fixed the PR based on other devs suggestions in https://github.com/DariusIII/nZEDb/commit/be3d488b8c833476631533f37276cf08b5f1defd

Issue #1880, opened 2015-09-19
Provide an option to download the TV Artwork to the TV Cover Folder instead of the SQL Database.


SchoolBusDriver:
Along with [Replacing TVRage (issue 1879)](https://github.com/nZEDb/nZEDb/issues/1879) with [thetvdb.com](http://thetvdb.com/) or [tvmaze.com](http://www.tvmaze.com/)

Provide an option to download the TV Artwork to the TV Cover Folder instead of the SQL Database. Movies Artwork does this now.

ruhllatio:
No longer optional... required as of commit 96bd66dc0f32a4b3defdaff1fc252c73cc3da735 :D

PR #1893, opened 2015-10-02
Moinax's TVDB PHP API Integration (Phase 1)

ruhllatio:
Integrates Moinax's API with a few namespace touch ups to match our conventions.  Includes a basic test script in misc/testing/Tests to prove out the API is working as it should based on the author's example.


SchoolBusDriver:
Impressive work!

Would be great if you could add the enhancement below to your TVDB PHP API Integration.

[[Issue 1880] Provide an option to download the TV Artwork to the TV Cover Folder instead of the SQL Database](https://github.com/nZEDb/nZEDb/issues/1880).

ruhllatio:
@SchoolBusDriver - not that impressive.  I'm basically just importing his library.  But, why reinvent the wheel as they say?  This is what open source projects are for!

Anyways, I have said time and time again that we need to move tvrage images out of the database.  Causes unnecessary bloat and is way less reliably IMO.  Almost all of the images I have for TV shows right now are corrupted when retrieving from the DB.  I will look into this when we revamp processing.

SchoolBusDriver:
> Anyways, I have said time and time again that we need to move tvrage images out of the database. Causes unnecessary bloat and is way less reliable IMO. 

Agreed!

> Almost all of the images I have for TV shows right now are corrupted when retrieving from the DB.

Agreed!

> I will look into this when we revamp processing.

Thanks!

updateTvRage.php and populate_tvrage.php files will need to be modified to accommodate the TVDB PHP API Integration. Once completed, updating the Templates, SQL functions, themes shouldn't too complicated since the everything else is written to the Covers already.

PR #1910, opened 2015-10-11
Move all the TV related classes so nzedb/processing/tv folder. Move the processTvReleases function into main TV class.

DariusIII:


niel:
Can't merge this as it is. The libs/Moinax/Trakt library should not be moved from the libs directory.

DariusIII:
I have updated the PR, but i have merged latest dev also.

PR #1915, opened 2015-10-14
Update DbUpdate.php

cauchy49:
when local, the sql should be "LOAD DATA LOCAL"
incorrect test : reverse it...

### DIFF
--- a/nzedb/db/DbUpdate.php
+++ b/nzedb/db/DbUpdate.php
@@ -98,7 +98,7 @@ class DbUpdate
 
 		$files = empty($options['files']) ? Misc::getDirFiles($options) : $options['files'];
 		natsort($files);
-		$local = $this->pdo->isLocalDb() ? '' : 'LOCAL ';
+		$local = $this->pdo->isLocalDb() ? 'LOCAL ' : '';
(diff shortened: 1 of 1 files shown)


ruhllatio:
@cauchy49 thanks for the PR, but we cannot accept merges to the master branch.  They have to go to the dev branch first.  Please re-open the PR and do a compare/pull request to the dev branch instead.

ruhllatio:
Wow, should have looked at it first too.  LOCAL is not actually for local DB servers.  It's to tell the MySQL client that the data file is local to the client while the server is remote.  Otherwise the Server searches for the file locally and fails to find said file.

cauchy49:
my sql server is local, and without this change I have a "permission denied" from mysql (permissions on directory afile are ok) when i try to configure the database (in step2.php). Maybe LOCAL must be always used...

ruhllatio:
Are you SURE you added the GRANT FILE rights?

niel:
if LOCAL were always used it would change how MySQL looks for the file and cause other problems. Additionally, extra permissions would be required by the user.

cauchy49:
yes, i try with root and nzedb account :

global permissions are  (from phpmyadmin) :

nzedb	localhost	Oui	SELECT, INSERT, UPDATE, DELETE, CREATE, DROP, __FILE__

niel:
Check all the directories in the path to your files have the Traversible bit set, and the files are readable by all.

Issue #1844, opened 2015-08-25
Not all "Covered" Movies showing ?


scottia:
I noticed my Movie counts (pages) with Covers slowly tapering down even though my Max Page Results is set at 500K.  However, if I manually change my "offset=XXXXX" number in the GUI (For example - http://192.168.0.8/movies?t=2000&title=&actors=&director=&rating=&genre=&year=&ob=&offset=14600#results) to a HIGHER number than what's shown in "Home" I see the missing Movie pages but the number count is off.

Here's a screen from a Manual edit.  Notice I have 760+ pages with covers (when manually changes) vs. 731 as shown in the 2nd screenshot below ?

![image](https://cloud.githubusercontent.com/assets/5298748/9457613/4d0f8cea-4aaa-11e5-94d4-cae7a4e13945.png)

Here is the 2nd screenshot showing a "Normal" home view of Movies with Covers.  Notice is shows 731 Movie releases with Covers vs. 760 show above?

![image](https://cloud.githubusercontent.com/assets/5298748/9457642/c4874542-4aaa-11e5-9b1f-c0c5e1e00ec3.png)

The started about 2 days ago when changes to Movies.php were made in DEV.  Since then I see the following error in my logs that may have something to do with it but I can't say for certain.

[24-Aug-2015 20:23:01 America/Chicago] PHP Notice:  Undefined index: title in /var/www/nZEDb/nzedb/Movie.php on line 1002

scottia:
Now I have dropped all the way down to 709 Covered movies showing.  Very odd ?

![image](https://cloud.githubusercontent.com/assets/5298748/9473104/9bc9f162-4b1e-11e5-977b-ed97699f9e01.png)

ruhllatio:
@scottia - is this still happening?  Are you still seeing those errors?  Which script?  I'm assuming movie processing...

ruhllatio:
Near as I can tell, that error is coming from the TraktTv response.  Don't know much about how important that is in the long run, but it seems we aren't checking the return value of 'title' from them.  If that's the only error, it's not hurting anything because it is only doing an echo.

scottia:
No error @ruhllatio 

The page count is off and slowely decreasing.  I am down to 702 pages of movies with covers according to the page but have over 750.  The longer I run, the lower it goes.  I don't appear to be missing releases..

ruhllatio:
[...]

PR #1920, opened 2015-10-22
Fix movie covers on alpha theme, add first batch of changes to omicro…

DariusIII:
…n and charisma templates for new tv changes


scottia:
you think this will fix my issue #1844 in Alpha ??

DariusIII:
No, this has nothing to do with your problem, it if purely the fix of the issue introduced by changes for tv processing.

ruhllatio:
@scottia - this is STILL a problem?  Balls.  Let me check if I actually pushed the fix for that query we discussed earlier.

scottia:
Oh yes, still an issue.  I can still click on the last page and add to the offset until I get to my "actual" last page.  I get the same results for books, audio, console & xxx as well.

It can wait until you finish dev-tv @ruhllatio 

Thanks again !

PR #1930, opened 2015-10-28
Remove totalrows count from functions, api and templates

DariusIII:


JudsonHat:
Why was this removed? It seems like useful information to have the total in the api results.

PR #1932, opened 2015-10-29
Add new codec indentifiers and consolidate old ones.

izolight:
Divx codec had a duplicate entry and i merged x264 and h.264 into one as the correct codec is h.264. I also added regexes for MPEG-4 (mostly pre xvid stuff) and hevc(h.265).

### DIFF
--- a/nzedb/ReleaseExtra.php
+++ b/nzedb/ReleaseExtra.php
@@ -27,23 +27,23 @@ class ReleaseExtra
 		if (preg_match('/XVID/i', $codec)) {
 			return 'XviD';
 		}
-		if (preg_match('/^27$/i', $codec)) {
-			return 'Blu-Ray';
-		}
-		if (preg_match('/V_MPEG4\/ISO\/AVC/i', $codec)) {
-			return 'x264';
-		}
 		if (preg_match('/wmv|WVC1/i', $codec)) {
 			return 'wmv';
 		}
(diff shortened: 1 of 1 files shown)


niel:
I may be misreading your PR, but it looks to me as though you have dropped x264 not merged it into h.264.
Please note, I've refactored this method in dev to use a switch instead of the multiple IFs. I've also changed the regex delimiter to '#' instead of '/' to make it more readable.
If you could either redo your PR accordingly, or just explain the above query I can merge it manually.

izolight:
the x264 and h264 are merged, as the regex also matches for "AVC" as there are more variants than just "/V_MPEG4\/ISO\/AVC/" which x264 matching used before.

niel:
Ok, I see what you mean, the check is less specific to catch more possibilities.
Could you explain this one to me:
((^20|FMP4|MP42|MP43|MPG4$)|ASP)
The anchor (^$) feel like they're in the wrong place. Shouldn't they be outside the enclosing parentheses? Like so:
(^(20|FMP4|MP42|MP43|MPG4)$|ASP)

izolight:
Oh yeah, no idea why i put them inside the parantheses.

niel:
Thanks. I'll make the changes manually.

niel:
Committed changes as 9e8d4e6, please take a look and tell me if they're what you intended.

izolight:
seems fine, thanks.

PR #1941, opened 2015-11-03
Fix for wrong redirect after release-edit

DariusIII:


niel:
Issue fixed with another method.

PR #1958, opened 2015-11-16
using rawurlencode instead of urlencode.

izolight:
Now shows with spaces can be matched as urlencode added '+' instead
of correct '%20' urlencoding. So the result was always empty.


Bart39:
Should this be pushed back to the source as well ? Otherwise we are deviating from standard

ruhllatio:
@Bart39 - When I have the time I plan to do a healthy PR to the base wrapper.

Bart39:
cool

PR #1959, opened 2015-11-17
Fixes for tvmaze lookup

izolight:
Lookup wasn't working for akas and also sometimes not returning a valid response.


izolight:
@ruhllatio 
You only got that since my commit right?
So checking if the tvmaze-aka is empty should do it.
https://github.com/izolight/nZEDb/tree/dev-izolight-tvmaze2

PR #1961, opened 2015-11-17
Prevent converting to utf8 multiple times

izolight:
if the input is already utf8 (which it is for many providers) a conversion with utf8_encode will result in garbled output if non-ascii characters are used.


Bart39:
this needs checking by someone more proficient than me :p

ruhllatio:
I'm seconding this.  It messes with the searchnames, and I'm not sure what kind of effects it will have.  @kevinlekiller @niel ?

izolight:
Anything with only ASCII Characters in the subject (probably 99% of posts) won't see any change. But posts with non ASCII Characters are at the moment wrongly converted. 
nzedb assummes that the input is iso-8859-1(http://php.net/manual/de/function.utf8-encode.php), but with most(all?) providers it is actually utf8. I tested this with astraweb, highwinds, xsnews, giganews and newsoo. They all give the result back in utf-8.
With ASCII Characters it doesn't matter as they are the same in both encodings.
But with other characters it converts them in a wrong way.
Example:
Original Subject: 
[KoreanTV] [Olive].비법.E02.150820.HDTV.H264.720p-WITH [1/31] - "[Olive].비법.E02.150820.HDTV.H264.720p-WITH.nfo" yEnc
with current nzedb it will get converted to this:
[KoreanTV] [Olive].ë¹ ë².E02.150820.HDTV.H264.720p-WITH [1/31] - "[Olive].ë¹ ë².E02.150820.HDTV.H264.720p-WITH.nfo" yEnc

kevinlekiller:
The issue with this is performance, when we download headers, we we iterate through tens of thousands of these in less than a second, each call of of Encoding::toUTF8 is 50x slower than a call to utf8_encode (which I know is causing issues, I'm just using to compare the performance penalty).

For example, on 50,000 headers this adds 1 to 2 seconds when using Encoding::toUTF8 vs 0.01 to 0.02 seconds with the utf8_encode just for the subject on a 4.00GHz haswell CPU, even more if you factor in the poster (probably another 0.5 to 1 second).

It's slow because it's iterating through every character and doing multiple conditionals, in PHP.

If we could use a PHP extension written in C or C++ that would do the same as this PHP library, it would not be as bad.

 Or we can not do anything, save the performance hit and the few that are encoded as iso-8859-1 will be messed up.

izolight:
Oh I didn't realize it was this much slower. So is there actually a provider that encodes the headers in iso-8859-1? If not then we could probably just remote all the conversion stuff and make the processing even faster.

ruhllatio:
What if we did mb_detect_encoding on the header first?  Then if we found multi-byte characters we run the lib?

kevinlekiller:
The Content-type field is supposed to list the encoding.

The Content-type field is not returned in the "OVER" command, which means we'd have to get the full header which is slower and not worth it.

You could check a few headers (using the HEAD command) to see what Content-type says, then you can see if most of them are UTF-8 and if it's worth the trouble trying to fix the issue.

kevinlekiller:
@ruhllatio "What if we did mb_detect_encoding on the header first? Then if we found multi-byte characters we run the lib?"

I just tried this, it's ~30% slower than a utf8_encode call (vs ~5000% for a Encoding::toUTF8 call). It's acceptable, I've tested some subjects, all of them were UTF-8 so far.

Example code:

`$subject = (mb_detect_encoding($subject) === "UTF-8" ? $subject : Encoding::toUTF8($subject));`

The mb_detect_encoding could possibly be added to the toUTF8 method then you wouldn't need to edit much of the PR.

kevinlekiller:
Actually, that last code is wrong, `mb_detect_order()` returns "ASCII" and "UTF-8", which means it was never checking for ISO-8859-1.

Adding a UTF-8 / ISO-8859-1 check to mb_detect_encoding() makes it 4x slower. Although it's still much faster than a call to Encoding:toUTF8

izolight:
The Content-Type header actually is just for the content of the body and for binaries not set anyway.
RFC3977 says that header values should only contain ascii values, and they say that utf-8 shouldn't be used. But it is used nowadays and everywhere the rfc states that everything else should be utf-8. 
So is there any reason at all to assume that the input is ISO-8859-1? I haven't found a provider that gives the headers in this format.

kevinlekiller:
I dunno, I was reading mainly the RFC's on email (usenet is based on email).

Mainly this https://www.ietf.org/rfc/rfc2047.txt

I think we should just assume it's ascii if that's what the standard says, saves us the head aches / performance issues.

izolight:
I was refering to section 10.2 of the NNTP standard (https://tools.ietf.org/html/rfc3977#section-10.2).
But if we assume it's ascii there is no need for the utf8_encode in the first place, or am i missing something?

kevinlekiller:
Yeah, it can be removed (utf8_encode), I think the utf8_encode was brought over from newznab, we haven't thought about it until now.

izolight:
I tested all three variants too (with update_binaries.php and 100000 headers, so the differences are a lot smaller).
toUTF8 was about 35-60% slower than no conversion and utf8_encode about 25-35% slower than no conversion. So removing this seems to bring a decent speedup too :)

kevinlekiller:
Nice, that's a huge difference.
I was toying some months ago with moving some of the code in that loop into a PHP extension but it would have required users to install the extension so I felt it wasn't worth the effort, I think the slowest part current is matching regex, especially for blacklists/whitelists.

izolight:
Has someone else besides me tested this so far? I haven't found any problems yet.

izolight:
I've got the following results for a normal groups (a.b.teevee), some minor speedups in processing collections:
a b teevee

And the results for a group with a lot of non-ascii characters (a.b.mm.japanese), Speedups are bigger:
a b mm jap

niel:
Looks promising. Unfortunately I'm not able to benchmark/test this myself.

niel:
I finally had time to look at this properly. 
@izolight  if you close this PR and resubmit it without the library addition/removal (i.e. simply removing the utf8_encode usage), I'll be happy to merge it into dev.

PR #1975, opened 2015-11-26
Dev themes

niel:


ruhllatio:
@niel Was this an oopsie or were you intending to merge something straight to master?

niel:
@DariusIII neither actually. As you can see this is marked NO-MERGE. I was only creating the branch for ease of pulling it into my test repo and the merge to get the cli commands to do it ;-)
I'll kill it to avoid any confusion though.

Issue #1966, opened 2015-11-19
tmux.conf needs updating 


bobtongue:
Tmux 2.1 has changed the syntax for handling the mouse. It needs to be updated in the tmux.conf. The way it is written right now, when you update to tmux 2.1, the mouse does not work in the tmux desktop. This would also require that you require tmux 2.1 be used. 

From the tmux changelog:
CHANGES FROM 2.0 to 2.1 18 October 2015

Incompatible Changes
====================

* Mouse-mode has been rewritten.  There's now no longer options for:
	- mouse-resize-pane
	- mouse-select-pane
	- mouse-select-window
	- mode-mouse

  Instead there is just one option:  'mouse' which turns on mouse support
  entirely.
* 'default-terminal' is now a session option.  Furthermore, if this is set
  to 'screen-*' then emulate what screen does.  If italics are wanted, this
  can be set to 'tmux' but this is still new and not necessarily supported
  on all platforms with older ncurses installs.
* The c0-* options for rate-limiting have been removed.  Instead, a backoff
  approach is used.

DariusIII:
Problem is that Tmux 2.1 has some dependencies not available on all systems, so forcing it to users would cause many problems.

niel:
I've already looked into this myself a month or two ago. The problem is, as you mention, the changes are incompatible with previous versions. Versions that many users are stuck with, until the distros update them.
If, and it's a big if, we can find a way to have tmux load different configuration files based on it's version we could do this before 2.x becomes the default across distributions.

Bart39:
if the mouse-mode is the only difference we can use if-shell in the conf file to check the tmux version and apply the correct commands

Bart39:
if-shell "[[ `tmux -V` == *2.1 ]]" 'command-to-be-run-here'

Bart39:
even better example using extra files for the version specific commands
if-shell "[[ tmux -V == *2.1 ]]" "source-file .tmux-21" "source-file .tmux-default"

niel:
That'll be fine until the version number changes. Is there a way to do > 2.0?

kaibosh:
As it's a *conf file, it could be moved into nzedb/config/ as tmux-example.conf. The example file could have separate lines for tmux 1.9, 2.x etc., to be activated at install time.

Practically it would also solve the feature request part of issue #1489 so that local tmux configs are not overwritten after a pull.

niel:
[...]

Issue #1489, opened 2014-09-04
Tmux with own config


NemesisRE:
First of all great Tool! Thank you for that.

I use tmux over a year now and my whole server is managed in tmux windows, the moment I connect on it it will open tmux with my own config.
Your Tmux script will not run if there is already a tmux running even if it is an other session and if i kill my session and run your tmux script then restart my own session it will use your config.

Would it be possible to have an option to use an own config and allow running other sessions when starting yours?

regards
Steven

niel:
Created a new branch on the repo (dev-tmux) with a start to dealing with these issues. Please let me know additional changes to make.

kaibosh:
Couple of things could be improved: 
 - set-remain-on-exit should be in the common .conf
 - set default-terminal is also in the common .conf. Not sure xterm-256color setting is needed for v2.1.
 - mouse-mode on/off could be copied to the 2.1 conf.

Should you be short on time, I can test these and submit a PR in the coming days.

niel:
@kaibosh Thanks. I'd appreciate the PR too. I'm working on some theme changes at the moment.

kaibosh:
Have the tmux v2.1 config basically finished. Playing with mouse scrolling. Should have PR ready tomorrow.

PR #1983, opened 2015-11-29
Updates for tmux v2.1 - Issue #1966

kaibosh:
Do not overwrite local tmux.conf - Issue #1489
Tested with v2.0 and v2.1 tmux. Needs testing with v1.9.
Powerline tmux config changes not tested, should work.
Unsure if existing tmux.conf files removed correctly from repo, my git-fu lacking here :)


niel:
@kaibosh could you do this against the dev-tmux branch already started, instead of against dev please.

kaibosh:
ok, will give that a go

PR #1985, opened 2015-11-29
Add Trakt support to Dev

b3rs3rk:
Would like to get this in after NM merge if possible, team.  Please review and comment.


nzedb/libraries/TraktAPI.php
```diff
@@ -12,14 +12,9 @@
 	const API_URL = 'https://api-v2launch.trakt.tv/';
 
 	/**
-	 * The Trakt.tv API v2 Client ID (SHA256 hash - 64 characters long string). Used for movie and tv lookups.
-	 * Create one here: https://trakt.tv/oauth/applications/new
-	 *
-	 * @var array|bool|string
+	 * @var array	List of site IDs that trakt,tv supports. Only trakt is guaranteed to exist.
```
niel: trakt,tv Typo?

niel:
Only other comment would be more adding to the DocBlocks would be nice ;-)

nzedb/libraries/TraktAPI.php
ruhllatio: Yep, typo.  Will fix before merge.

ruhllatio:
For future reference @niel What additional information would you like to see?

niel:
@ruhllatio Just basic description of what the method does, what each parameter represents, and any return value's meaning/use.
BTW this is something we all need to do better at.

ruhllatio:
Okay.  I will keep that in mind for future work.

PR #1988, opened 2015-12-02
Dev themes

niel:


ruhllatio:
I pulled this down to my dev machine.  Was running alpha during the pull.  When I first connected to the site, it showed the default theme.  When I went to Site Edit it was set to Alpha.  I saved it and returned to the main site and the CSS is all jacked up.  Missing some background colors.  The cart and send to queue icons when browsing releases are all Download NZB icons.  New poster wall is missing all icons for each item in all categories.  Just too much missing in Alpha.

ruhllatio:
http://pastebin.com/yNsHK7Uk

niel:
ok, will look into that.
No idea why Alpha would be affected so badly, it wasn't exactly changed.

ruhllatio:
If you'll notice, all of the 404s are to themes/alpha not themes/Alpha.  You capitalized the A in your merge so it broke the URLs to these files.

niel:
yeah, I thought the path would be dynamically created by the themes. I'll make sure that it is for them all.

ruhllatio:
Sounds like a plan.  If you made it respective to the theme, they could be moved at will down the road if we had to and wouldn't have to worry about situations like that.

niel:
Not sure what you mean by 'respective to the theme.'?
I was thinking to create a variable which would contain 'themes/<theme-name>' and use that in any paths.

www/admin/site-edit.php
```diff
@@ -211,10 +211,11 @@
 $themelist = [];
 $themes    = scandir(nZEDb_WWW . "/themes");
 foreach ($themes as $theme) {
-	if (strpos($theme, ".") === false && is_dir(nZEDb_WWW . "/themes/" . $theme)) {
+	if (strpos($theme, ".") === false && is_dir(nZEDb_THEMES . $theme) && ucfirst($theme) === $theme) {
```
kevinlekiller: `ucfirst($theme) === $theme`

This should be reverted, since this can cause issues with people using custom themes. It's undocumented and nothing is displayed to hint people with custom themes why their theme is not loading.

niel: I don't see a need to revert it, unless you can suggest a better alternative.
It *will* be documented in ChangeLog and ReleaseNotes. It is true nothing would be displayed to say why themes might stop, but checking the ChangeLog/ReleaseNotes should quickly solve that if they hadn't read them before hand, as any site admin should.
I chose that method as it made semantic sense to me - themes are named, and as proper nouns they should have a capital first letter. The additional intent is that we can use lower-cased names as internal themes. So 'admin' can be moved to be a theme and not part of shared, which seems weird to me.
In the long run, I think this will make themes more consistent and easier for all.

kevinlekiller:
Another reason, the theme name is stored in the DB ([per user also](https://github.com/nZEDb/nZEDb/blob/31a5543abe9d2d463776cba3344f95230d8b982b/www/pages/profileedit.php#L111), but you didn't add the capitalization check there), so if someone is using the current "alpha" theme, and they pull this change, the website will not load properly or at all because in the DB it's still "alpha", but it's not available in that array, "Alpha" is.

niel:
Wasn't aware of that, I'll supply an SQL patch to fix that, if we decide to continue this path.

kevinlekiller:
I think that's what ruhllatio meant by the 404 errors, but wasn't sure.

niel:
I don't think so, at least I found some hard-coded paths in basepage.tpl so assumed they're the main cause. Seems to apply to most (if not all) of the themes, I assume they got copy/pasted and spread the use.

DariusIII:
I have implemented the same method in nntmux, and there is no need to revert anything. If the template is not found (in per user selected template case) it will simply revert to site default one and user will just have to select the template again. Errors ruhlattio mentioned are because of the hardcoded paths in scripts we are using.

PR #1990, opened 2015-12-04
Dev tmux

niel:
This looks good to me, but it needs testing by people using tmux < v2.1


kaibosh:
Been testing this branch today and came across issue where tmux v2.1 crashes after about 10 mins. Tried the latest git version (v2.2) and seems to be worse. v2.0 runs fine. Still playing with various options looking for a stable config. Is v2.1 stable for others with this config?

ruhllatio:
I'm getting this:


Notice: Undefined index: error in /var/www/nZEDb/libs/smarty/templates_c/0cecb05e6da0328118c9b0ba1bd28c74172f5452_0.file.tmux-edit.tpl.php on line 93

Notice: Trying to get property of non-object in /var/www/nZEDb/libs/smarty/templates_c/0cecb05e6da0328118c9b0ba1bd28c74172f5452_0.file.tmux-edit.tpl.php on line 93

On the tmux settings page in Admin.  Additionally, I can't get tmux to start using 14.04's default of Tmux 1.8.

ruhllatio:
+if-shell "[[ `tmux -V` > 'tmux 2.0' ]]" "source-file tmux-2.1.conf" "source-file tmux-2.0.conf"

This appears to be creating a file called tmux 2.0 in my tmux directory.  I thought it was intended to be an evaluation.

kaibosh:
Tried v1.8 and it fails here also. The if-shell test borks it somehow. Not seeing it create another file.

kaibosh:
Status of the testing I've done with these configs:
tmux 1.8 -> tmux 2.0 work fine as before.
tmux 2.1 & 2.2(git checkout): nzedb runs ok for about 10 mins, then the tmux server session crashes randomly. Cannot find anything obvious with tmux -vvvv debugging. Using strace is difficult as there's so much running.
Have tried stripping the tmux.conf file down to the only required param for nzedb (set-remain-on-exit) and tmux v2.1 still crashes. (i.e. with no mouse mode active).
Currently suspect one of the tmux clients sending high ASCII values (e.g. escaped colour codes) might be causing the tmux server process to crash. Or the default-terminal setting screen|xterm-256color with/without UTF8, have tried various combinations to no avail. Possibly could be distro/environment related as this testing done on latest fedora which tends to be bleeding edge.
This branch could be merged as far as I can see it does not make the situation worse. Also having a local tmux.conf would allow easier testing by others once tmux v2.1 rolls out.
Compiling tmux v2.0 from source is easy so documenting that as a work-around is an option until this gets sorted.

niel:
@kaibosh thanks for the update. I'll talk with other devs about merging into dev.

scottia:
Any chance of this being merged to Dev ?

niel:
@scottia Yes, next few days. Has been on my to do list for a while.

kaibosh:
Please don't merge these old changes. The tmux config files merged into dev last week are better than those in this PR.

niel:
Damnit 
@kaibosh  weren't these the updated ones? Would you check these actual files, because I merged your recent changes into this branch before merging into dev.

kaibosh:
Heh, I got a bit nervous as this is the old PR from Dec15. The newer PR 2321 was merged 5 days ago from my own branch.

I've checked the primary dev branch and everything looks OK (if I can believe the github web interface hasn't cached something).

I'll double check tomorrow with a proper git pull as I've got a small update to start.php to submit anyway.

kaibosh:
@niel for some weird reason it looks like this merge deleted the tmux.conf file in the powerline directory and placed my older tmux.conf* files there instead.

As a quick fix, could we get the current 0.x powerline/tmux.conf back in the powerline directory?

In the coming days I'll look to split that config also for the various tmux versions.

niel:
@kaibosh ok, I'll work on reverting it today.

PR #2000, opened 2015-12-15
Add full movie view when requesting imdb from movies page

DariusIII:


ruhllatio:
I'm guessing there is more to be committed?  A cursory check shows that viewmoviefull.tpl doesn't exist in Alpha.  So this would likely cause breakage.

scottia:
Nice work on Gamma.. Little thing.  Clicking cover OR cover button has mixed results on what's displayed.  Also no Movie description on some.

www/themes/Gamma/templates/frontend/profileedit.tpl
```diff
@@ -55,9 +55,7 @@
 		<div class="control-group">
 			<label class="control-label">Change site theme</label>
 			<div class="controls">
-				{if $page->settings->getSetting('userselstyle') == 1}
-					{html_options id="style" name='style' values=$themelist output=$themelist selected=$user.style}
-				{/if}
+				{html_options id="style" name='style' values=$themelist output=$themelist selected=$user.style}
```
niel: The profile.tpl needs to display this too.

DariusIII: Pushed the suggested changes.

www/themes/Gamma/templates/frontend/profile.tpl
```diff
@@ -54,7 +58,7 @@
 		<th>Grabs Total:</th>
 		<td>{$user.grabs}</td>
 	</tr>
-	{if ($user.id==$userdata.id || $userdata.role==2) && $site->registerstatus==1}
+	{if (!$publicview || $isadmin) && $site->registerstatus==1}
```
niel: I'm not sure about the !$publicview. I'm reading this as if NOT public or is admin user, which doesn't make sense to me. Shouldn't it be if IS public or IS admin user?

www/themes/charisma/templates/frontend/profile.tpl
```diff
@@ -105,7 +127,7 @@
 															{/if}
 															</tbody>
 														</table>
-														{if ($user.id==$userdata.id || $isadmin) && $site->registerstatus==1}
+														{if (!$publicview || $isadmin) && $site->registerstatus==1}
```
niel: Same as for Gamma version of this file.

www/themes/omicron/templates/frontend/profile.tpl
```diff
@@ -105,7 +127,7 @@
 															{/if}
 															</tbody>
 														</table>
-														{if ($user.id==$userdata.id || $isadmin) && $site->registerstatus==1}
+														{if (!$publicview || $isadmin) && $site->registerstatus==1}
```
niel: Same as for Gamma version of this file.

www/themes/Gamma/templates/frontend/profile.tpl
DariusIII: https://github.com/nZEDb/nZEDb/blob/master/www/themes/alpha/templates/frontend/profile.tpl#L44

It is same as in alpha.

www/themes/charisma/templates/frontend/profile.tpl
DariusIII: It is already fixed in newznab-tmux, just did not update files for this PR.

www/themes/omicron/templates/frontend/profile.tpl
DariusIII: It is already fixed in newznab-tmux, just did not update files for this PR.

www/themes/charisma/templates/frontend/profile.tpl
DariusIII: Same comment as for Gamma :)

www/themes/omicron/templates/frontend/profile.tpl
DariusIII: Same comment as for Gamma :)

www/themes/Gamma/templates/frontend/profile.tpl
niel: ok. I'll merge it as is, but I'm still unsure about it. May be alpha is wrong too ;-)

PR #2004, opened 2015-12-20
Change wrong video sample extension used in Gamma.

DariusIII:


DariusIII:
Will do a PR against dev-themes.

PR #2012, opened 2015-12-31
Fixed paths in themes

egandt:
Fixed paths for 2 themes that were defaulting to the root /


niel:
Good catch, I'll merge  them in. Thanks.

PR #2020, opened 2016-01-06
Add Limited SRR Support for Release Name Fixing

b3rs3rk:


Fossil01:
Could we also store CRC hashes of files inside the RAR's or pull them from the SRR is possible?

rarinfo should be able to do this easily. I already run an API with a fairly large CRC database, trying to get automated updates now as well. This could hugely help with renaming.

ruhllatio:
@Fossil01 - Possible, yes.  Let's talk in IRC when you have a chance.

niel:
The use of an array constant is only allowed in PHP 5.6+, so we came up with an alternative to use until 5.5 reaches its EOL in July. See Category::getCategoryOthersGroup() in dev branch.

Fossil01:
Could you do something with my API @ruhllatio ? It's syncing with SRRDB.com every hour for all releases bigger than 100MB. It can search for CRC32 & filesize to get an exact match and it can search OSO hashes if available.

API is fully functional, if It's used by nZEDB I just need to implement some sort of API key system for different users.

ruhllatio:
@Fossil01 - This wasn't the best implementation.  Rethinking it along the lines of CRC.  Mocking it up now.  Will let you know when I have it.

ghost:
Roger that.

PR #2023, opened 2016-01-08
Fix wrong path to shared folder in Gamma theme.

DariusIII:


DariusIII:
I have fixed some issues in Charisma/Gamma/Omicron templates.

niel:
I think these are already fixed in dev-themes.

DariusIII:
Some are, but not all of them.

PR #2030, opened 2016-01-12
Fix download basket in Gamma, fix wrong file extension preg_match in NZB class.

DariusIII:


DariusIII:
I have updated this PR with low res RT images, as they are the only thing that is pushed via RT after their proxy change.

scottia:
Thank you for taking the time to make these changes!!  I for one appreciate it

niel:
Damn. missed this PR originally because of IRC problems. I'll merge it in later today.

PR #2037, opened 2016-01-18
Move Other cats to 0000 range, books cat to 7000 range.

DariusIII:


Wally73:
dont see a reason for this just move the category other stuff all to all other (7000) think then less scripts etc has to be changed (but ok that is my opinion)

DariusIII:
It is more than a move, there is no logic for misc cats to be between xxx and books, and moving books to 7k range is a logical step.

DariusIII:
I have made a change so Books class will use constants from Category instead of numbers for category ids.

niel:
I thought you were going to change all uses of numbers in code to use constants?

DariusIII:
I changed to constants where i could. The rest of the code is appending stuff to queries to build them, and i am not familiar in a way to make it use constants, and to not break those.

DariusIII:
With help @Bart39 i have fixed the rest of the queries and updated the PR.

niel:
Sweet. I'll look at it ASAP

PR #2041, opened 2016-01-22
Update Noty package to version 2.3.8 and move it to shared/libs folder

DariusIII:


niel:
Can you add a subdirectory in shared location i.e. .../www/themes/shared/libs/noty/v2/ like I did with bootstrap. Just in case we need to have different version at some point.

DariusIII:
There is no need for versioning folders, as noty is an universal package. Situation is not similar to bootstrap, where diferences in versions can be a show stoppers.

niel:
So version 1 would work in there just as well as version 2? Or are you saying that each version is a superset of the previous major number and nothing is removed?

DariusIII:
Version 1.x is very old as is not used in project. And yes for the second part.

niel:
ok, that's what I needed to know. I had it on my list of things to find out so I could move it to shared, seems you have it all covered ;-)
I'll merge in over the weekend at some point.

PR #2078, opened 2016-03-05
Add a trailing slash to sab url if one does not exist.

zetas:
Further functions expect a trailing slash and one is not enforced in profileedit. I considered enforcing it there but then I'd have to write a script to update everyones sab url info in the database (or use this code as well), this way I achieve the necessary tweak with 2 lines of code total.


niel:
Fixed in dev

PR #2079, opened 2016-03-07
Prevent converting to utf8 multiple times

izolight:
Clean version of this PR https://github.com/nZEDb/nZEDb/pull/1961


niel:
Thanks.

niel:
We had to revert this from the repo as it was causing INSERT errors for every row on more than one person's setup.
Can you verify it and let me know the result.

izolight:
i will try to reproduce the error

PR #2113, opened 2016-04-19
Create simple-expanded

Wally73:
Addresses issue #. none its an expanding of simple.sh script in screen/sequential
## 
## 

this is my simple-expanded script (with stuff in it from an old threaded.sh script) but for the ones which aren't commented out work fine on my box 

there are some obsolete lines in it (but those are commented # so aren't used anyway)

feel free to modify it in anyway you see fit (and check if paths are right on your system)


niel:
@Wally73 Thanks. Looks good to me, I'll merge it in to dev and fix the one 'problem' I see.

PR #2120, opened 2016-04-27
Check for MovieSD before MovieHD in Categorize.php

ghost:
Need to check for MovieSD before MovieHD since MovieHD can be true even though it's actually a SD release. This is because of the "x264" in the regex. Not sure why it's there but this is an ez fix for it.


niel:
Please see our contributing guidelines for how we accept PRs. https://github.com/nZEDb/nZEDb/wiki/Contributing-Code-For-Beginners-to-Git(hub)
Main thing to note is we expect PRs to be in a branch (not from dev). I'll manually update the code for this as it is such a simple change.
Thanks for your work. ;-)

ghost:
Sorry, I'll try to do better in the future.

PR #2122, opened 2016-04-29
Delete Releases with 0 completion

daimoniac:
Addresses issue #.
## 
##


kevinlekiller:
I don't think this is the right way to do it, because 0 can mean the completion has not been checked yet.

Bart39:
i agree, this is not the right way to do it

Bart39:
closing this as the dev's have agreed on a solution - will create a separate issue

Issue #1581, opened 2015-01-18
Req Feature: Search Movies


Loopher:
On the Admin page, under the option -> Release List -> Movies -> View (/admin/movie-list.php).  I would like to request a search option be added here to find a movie title that has already been scraped from the IMDB or Movie DB sites.  I have encountered some entries that are missing a poster or background (fanart) and it is very cumbersome to search for these titles by clicking on each page.

Another thing that I would like to request is a rescrape option for this section.  Some of the titles may have been prematurely added before the movie posters/fanart have been added to the IMDB/MovieDB sites.  

I really appreciate all that you guys do.  Keep up the awesome work!

Loopher

PR #2125, opened 2016-04-29
Req Feature: Search Movies #1581

ghost:
Added feature request #1581


b3rs3rk:
It looked fine except you didn't actually pass $movietitle to the getRange function in movie-list.php.  Once I added that the search worked as expected.  I'm guessing it was a change you forgot to push to your branch.  Thanks for the PR!

PR #2134, opened 2016-05-07
Remove annoying NZBs

ghost:
Usenet uploaders often include their own NZB. This gets its own release in the release table unless you change so you have more than 1 file required and what not. but if you don't want that you have annoying NZB releases cluttering up things. I added an option to hide these while browsing and instead be able to download the NZB release in the details view. Only included changes to theme Alpha because that's the one I use.

### DIFF
--- a/nzedb/Releases.php
+++ b/nzedb/Releases.php
@@ -62,7 +62,8 @@ class Releases
 		$this->passwordStatus = ($this->pdo->getSetting('checkpasswordedrar') == 1 ? -1 : 0);
 		$this->sphinxSearch = new SphinxSearch();
 		$this->releaseSearch = new ReleaseSearch($this->pdo, $this->sphinxSearch);
-		$this->showPasswords = self::showPasswords($this->pdo);
+        $this->showPasswords = self::showPasswords($this->pdo);
+        $this->hideuploadednzb = ($this->pdo->getSetting('hideuploadednzb') == 1 ? true : false);
 	}
 
 	/**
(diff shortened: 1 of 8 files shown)


resources/db/schema/data/10-settings.tsv
```diff
@@ -21,6 +21,7 @@ archive	fetch	end	0	Download/process the last rar or zip file(s) to check for a
 		fixnamesperrun	10	The maximum number of releases to check per run(threaded script only).	fixnamesperrun
 		fixnamethreads	1	The number of threads for fixReleasesNames. This includes md5, nfos and filenames.	fixnamethreads
 		grabstatus	1	Whether to update download counts when someone downloads a release.	grabstatus
+        hideuploadednzb 0   Hide NZB files that were uploaded to usenet by the original poster. hideuplaodednzb
```
b3rs3rk, line 24: Needs indentation fix here.

www/pages/details.php
```diff
@@ -101,7 +101,8 @@
 		'privateprofiles' => ($page->settings->getSetting('privateprofiles') == 1 ? true : false),
 		'releasefiles'    => (new ReleaseFiles($page->settings))->get($data['id']),
 		'searchname'      => $releases->getSimilarName($data['searchname']),
-		'failed'          => $fail->getFailedCount($data['id']),
+        'failed'          => $fail->getFailedCount($data['id']),
```
b3rs3rk, line 104: Fix indentations here.

nzedb/Releases.php
```diff
@@ -1190,6 +1235,32 @@ public function getSimilarName($name)
 		return implode(' ', array_slice(str_word_count(str_replace(['.', '_'], ' ', $name), 2), 0, 2));
 	}
 
+   /**
```
b3rs3rk, line 1238: Would need to comment what this function does.

nzedb/Releases.php
```diff
@@ -199,7 +200,50 @@ public function getBrowseRange($cat, $start, $num, $orderBy, $maxAge = -1, $excl
 	{
 		$orderBy = $this->getBrowseOrder($orderBy);
 
-		$qry = sprintf(
+        if ($this->hideuploadednzb) {
```
b3rs3rk, line 203: Duplicating the entire query is bad.  Just add an additional %s to the existing query and do the substitution in the correct place like this:

```
 +                FROM
+                (
+                    SELECT r.*, g.name AS group_name
+                    FROM releases r
+                    LEFT JOIN groups g ON g.id = r.group_id
+                    WHERE r.nzbstatus = %d
+                    %s
+                    AND r.passwordstatus %s
+                    %s %s %s %s
+                    ORDER BY %s %s %s
+                ) r

+                ORDER BY %7\$s %8\$s",
+                NZB::NZB_ADDED,
+                (isset($this->hideuploadednzb) ? 'AND r.name NOT LIKE "%.nzb\" yEnc" : ''),
+                $this->showPasswords,
+                $this->categorySQL($cat),
+                ($maxAge > 0 ? (" AND postdate > NOW() - INTERVAL " . $maxAge . ' DAY ') : ''),
```

nzedb/Releases.php
```diff
@@ -199,7 +200,50 @@ public function getBrowseRange($cat, $start, $num, $orderBy, $maxAge = -1, $excl
...
+                    re.releaseid AS reid,
+                    v.tvdb, v.trakt, v.tvrage, v.tvmaze, v.imdb, v.tmdb,
+                    tve.title, tve.firstaired
+                FROM
+                (
+                    SELECT r.*, g.name AS group_name
+                    FROM releases r
+                    LEFT JOIN groups g ON g.id = r.group_id
+                    WHERE r.nzbstatus = %d
+                    AND r.name NOT LIKE '%%.nzb\" yEnc'
```
b3rs3rk, line 219: This may not be the best choice of filtering.  Doing any kind of LIKE may be fine with 100k releases.  But some of the bigger guys have millions of releases.

kevinlekiller:
It's probably better to filter this using blacklists or something.

ghost:
using blacklist removes the releases. which I don't want. Unless anyone else than me actually cares about having access to uploaded companion nzbs I will remove this pull request rather than fix the issues with it.

PR #2147, opened 2016-05-16
Feature composer asset plugin

DariusIII:
Addresses issue #.

None

##


niel:
After leaving it for a while and thinking it over, I'm convinced ALL assets handled by this change, should have an included version number for their top directory. We've already had an instance of a non-versions asset being updated and causing problems (fontawesome). We should pre-emptively prevent this from happening again.

PR #2150, opened 2016-05-18
feature db groups.id

niel:
- Update tables/code (except TPG tables) to use groups_id instead of group_id


misc/testing/Release/resetSearchname.php
```diff
@@ -19,7 +19,7 @@
 }
 
 if (isset($argv[1]) && $argv[1] == "full") {
-	$res = $pdo->query("SELECT releases.id, releases.name, releases.fromname, releases.size, groups.name AS gname FROM releases INNER JOIN groups ON releases.group_id = groups.id");
+	$res = $pdo->query("SELECT releases.id, releases.name, releases.fromname, releases.size, groups.name AS gname FROM releases INNER JOIN groups ON releases.groups_id = groups.id");
```
ruhllatio: We should probably standardize the formatting of these and use table aliases instead of table.column format.

resources/db/schema/mysql-ddl.sql
```diff
@@ -460,14 +460,13 @@ DROP TABLE IF EXISTS missed_parts;
 CREATE TABLE missed_parts (
   id       INT(16) UNSIGNED NOT NULL AUTO_INCREMENT,
   numberid BIGINT UNSIGNED  NOT NULL,
-  group_id INT(11) UNSIGNED NOT NULL DEFAULT '0'
-  COMMENT 'FK to groups',
+  group_id INT(11) UNSIGNED NOT NULL DEFAULT '0' COMMENT 'FK to groups.id',
```
b3rs3rk, line 463: Missed parts is a TPG table as well.  Probably don't want to change this if you're also not changing C/B/P.

niel, line 463: Only the line formatting and comment is changed, it's still group_id not groups_id. I intend to do all tpg tables as a separate patch, as it needs to handle both tpg and non-tpg cases. We should just make TPG the default ;-)

b3rs3rk, line 463: Ugh, you're right.  Scrolling blindness.

PR #2153, opened 2016-05-24
Feature media

b3rs3rk:
Addresses no issues as it is a feature enhancement.

- Add unique mediainfo ID matching


resources/db/patches/mysql/+2~releases.sql
```diff
@@ -1,3 +1,3 @@
 # This patch will add the proc_uid column to the releases table
 
-ALTER TABLE releases ADD COLUMN proc_uid TINYINT(1) NOT NULL DEFAULT '0' COMMENT 'Has the release been UID processed?' AFTER proc_srr;
+ALTER TABLE releases ADD COLUMN proc_uid TINYINT(1) NOT NULL DEFAULT '0' COMMENT 'Has the release been UID processed?';
```
b3rs3rk: This should be specified with an AFTER.  Since dev_srr isn't in Zed, you should select the next available column from our schema in case others have additional columns.  proc_files is the last one.

resources/db/patches/mysql/+2~releases.sql
```diff
@@ -0,0 +1,3 @@
+# This patch will add the proc_uid column to the releases table
+
+ALTER TABLE releases ADD COLUMN proc_uid TINYINT(1) NOT NULL DEFAULT '0' COMMENT 'Has the release been UID processed?';
```
b3rs3rk: AFTER proc_files

resources/db/schema/mysql-ddl.sql
```diff
@@ -693,6 +693,7 @@ CREATE TABLE         releases (
   proc_par2         TINYINT(1)                     NOT NULL DEFAULT '0',
   proc_nfo          TINYINT(1)                     NOT NULL DEFAULT '0',
   proc_files        TINYINT(1)                     NOT NULL DEFAULT '0',
+  proc_uid        TINYINT(1)                     NOT NULL DEFAULT '0',
```
b3rs3rk: Fix spacing

resources/db/schema/mysql-ddl.sql
```diff
@@ -756,6 +757,17 @@ CREATE TABLE release_comments (
   COLLATE = utf8_unicode_ci
   AUTO_INCREMENT = 1;
 
+  DROP TABLE IF EXISTS release_unique;
+  CREATE TABLE release_unique (
+  releases_id   INT(11) UNSIGNED  NOT NULL COMMENT 'FK to releases.id.',
+  uniqueid BINARY(16)  NOT NULL DEFAULT '0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0' COMMENT 'Unique_ID from mediainfo.',
+  PRIMARY KEY (releases_id),
+  INDEX ix_release_unique_uniqueid(uniqueid)
```
b3rs3rk: Remove the additional index and expand the PRIMARY KEY across releases_id, uniqueid.  In a two column table, you should expand the PK over the two columns if you also intend to add a secondary index over the second column.  This way, both columns are available in a PK lookup and provide additional performance in JOINs with the JOIN on the first column and the WHERE constraint(s) on the second column.  Doing it this way has no detriments as we are not wasting any additional space and only has performance benefits.  This is especially true for MyISAM users as MyISAM does not cluster the PK with the secondary keys.

nzedb/libraries/Forking.php
```diff
@@ -519,6 +520,11 @@ private function fixRelNamesMainMethod()
 				$where = "r.proc_files = 0";
 				break;
 
+			case "uid":
+				$join = "STRAIGHT_JOIN release_unique ru ON ru.releases_id = r.id";
```
b3rs3rk: This can probably be a LEFT JOIN with a WHERE ru.releases_id IS NOT NULL.  STRAIGHT_JOIN is a last resort when the LEFT JOINed table is still being used as the first table in the query.  With it being at best 1-to-1 with releases I'd say this won't fall into the same category like release_files.

nzedb/ReleaseExtra.php
```diff
@@ -210,6 +216,15 @@ public function addSubs($releaseID, $subsID, $subslanguage)
 		}
 	}
 
+	/**
+	 * @param $releaseID
+	 * @param $uniqueid
+	 */
+	public function addUID($releaseID, $uniqueid)
+	{
+		$this->pdo->queryExec(sprintf("INSERT IGNORE INTO release_unique (releases_id, uniqueid) VALUES (%s, UNHEX('%s'))", $releaseID, $uniqueid));
```
b3rs3rk: We are moving away from sprintf() arguments wherever possible in favor of double quoting the query string and inserting the values directly.

In this case, drop the sprintf and replace your %s with {$releaseID} and {$unqiueid} respectively.  Then look at how much cleaner PHPStorm says the query is when you're done.

``` php
$this->pdo->queryExec("
   INSERT IGNORE INTO release_unique (releases_id, uniqueid)
   VALUES ({$releaseID}, UNHEX('{$uniqueid}'))"
);
```

Additionally, be careful about using IGNORE.  While it won't matter much on single line inserts with a low traffic table, doing an INSERT IGNORE instead of a dupe check SELECT before hand can cause locks for no reason especially when running this routine threaded.

nzedb/NameFixer.php
```diff
@@ -374,32 +376,105 @@ public function fixNamesWithPar2($time, $echo, $cats, $nameStatus, $show, $nntp)
...
+			$total = $releases->rowCount();
+			if ($total > 0) {
+				$this->_totalReleases = $total;
+				if ($guid === '') {
+					echo $this->pdo->log->primary(number_format($total) . ' unique ids to process.');
+				}
+				foreach ($releases as $rel) {
+					$this->checked++;
+					$this->done = $this->matched = false;
+					$this->checkName($rel, $echo, $type, $nameStatus, $show, $preId);
```
b3rs3rk: Call uidCheck directly here.  We have probably already run checkName on this release before.

nzedb/NameFixer.php
```diff
@@ -1494,4 +1578,38 @@ public function fileCheck($release, $echo, $type, $namestatus, $show)
...
+	 * @param         $release
+	 * @param boolean $echo
+	 * @param string  $type
+	 * @param         $namestatus
+	 * @param         $show
+	 */
+	public function uidCheck($release, $echo, $type, $namestatus, $show)
+	{
+		if ($this->done === false && $this->relid !== $release["releases_id"]) {
+			$result = $this->pdo->queryExec(
```
b3rs3rk: You are selecting almost every release that has a uniqueid every single iteration.  Point the query against the release_uniqueid table instead and join releases to it once you've found a Pre match by the uid like this:

``` php
            $result = $this->pdo->queryOneRow("
                        SELECT r.id AS releases_id, r.size AS relsize, r.name AS textstring, r.searchname AS searchname, r.predb_id, HEX(ru.uniqueid) AS uid
                        FROM release_unique ru ON ru.releases_id = r.id
                        LEFT JOIN release_unique ru ON (ru.releases_id = r.id)
                        WHERE r.predb_id > 0 AND ru.uniqueid = UNHEX('{$release['uid']}') AND ru.releases_id != {$release['releases_id']}"
);
```

This way we are only looking up and retrieving a single row.

nzedb/NameFixer.php
```diff
@@ -1494,4 +1578,38 @@ public function fileCheck($release, $echo, $type, $namestatus, $show)
...
+				sprintf('
+						SELECT r.id AS releases_id, r.size AS relsize, r.name AS textstring, r.searchname AS searchname, r.predb_id, HEX(ru.uniqueid) AS uid
+						FROM releases r
+						INNER JOIN release_unique ru ON (ru.releases_id = r.id)
+						WHERE (r.isrenamed = %s OR r.categories_id NOT IN(%s)) AND r.predb_id > 0',
+						self::IS_RENAMED_DONE,
+						$this->othercats
+						)
+					);
+			foreach ($result as $rel) {
```
b3rs3rk: After changing the query above, we no longer need to foreach anything.  Either we find a match and run updateRelease with the required parameters or we return false for no match.

nzedb/NameFixer.php
```diff
@@ -1494,4 +1578,38 @@ public function fileCheck($release, $echo, $type, $namestatus, $show)
...
+				$percentage = $rel['relsize'] * (5/100);
```
b3rs3rk: A PreDB match on a unique ID of any kind is good enough.  Remove the size calculations or move them to the query above.

nzedb/NameFixer.php
```diff
@@ -1494,4 +1578,38 @@ public function fileCheck($release, $echo, $type, $namestatus, $show)
...
+				$minSize = $rel['relsize'] - $percentage;
+				$maxSize = $rel['relsize'] + $percentage;
+				if ($rel['uid'] === $release['uid'] && ($minSize <= $release['relsize'] && $maxSize >= $release['relsize'])) {
```
b3rs3rk: After removing the foreach above, this should be the only if conditional left after the query.  Simply say:

``` php
if ($result !== false) {
   $this->updateRelease($release, $rel['searchname'], $method = "uidCheck: Unique_ID", $echo, $type, $namestatus, $show, $rel['predb_id']);
}
```

resources/db/patches/mysql/+1~release_unique.sql
```diff
@@ -0,0 +1,10 @@
+  DROP TABLE IF EXISTS release_unique;
+  CREATE TABLE release_unique (
+  releases_id   INT(11) UNSIGNED  NOT NULL COMMENT 'FK to releases.id.',
+  uniqueid BINARY(16)  NOT NULL DEFAULT '0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0' COMMENT 'Unique_ID from mediainfo.',
+  PRIMARY KEY (releases_id),
+  INDEX ix_release_unique_uniqueid(uniqueid)
```
b3rs3rk: Remove the additional index and expand the PRIMARY KEY across releases_id, uniqueid.  In a two column table, you should expand the PK over the two columns if you also intend to add a secondary index over the second column.  This way, both columns are available in a PK lookup and provide additional performance in JOINs with the JOIN on the first column and the WHERE constraint(s) on the second column.  Doing it this way has no detriments as we are not wasting any additional space and only has performance benefits.  This is especially true for MyISAM users as MyISAM does not cluster the PK with the secondary keys.

PR #2169, opened 2016-06-09
Dev fixrelsinglerun

b3rs3rk:
Addresses poor query performance and high MySQL CPU Time caused by fixReleaseNames queries.

- fixRelNames multiprocessing no longer runs individual processes for each rename type.  One work query is done to find all eligible releases (per thread) and eligible releases are are tried against each work type in order of match preference.
- In order to streamline post processing work flow, renames now ONLY occur AFTER post processing and NFO processing has already completed to minimize duplication of processing effort and maximize rename probability.
- Tmux monitor pane tracks releases eligible for renames and triggers pane spawn to reflect this queue.


DariusIII:
Works fine. As far as i am concerned it can be merged.

PR #2177, opened 2016-06-15
Update IRCScraper.php

Wally73:
Addresses issue #.
## 
## 

"force" join of both channels

### DIFF
--- a/nzedb/IRCScraper.php
+++ b/nzedb/IRCScraper.php
@@ -157,7 +157,7 @@ class IRCScraper extends IRCClient
 		}
 
 		// Join channels.
-		$channels = defined('SCRAPE_IRC_CHANNELS') ? unserialize(SCRAPE_IRC_CHANNELS) : ['#nZEDbPRE' => null];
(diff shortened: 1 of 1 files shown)


DariusIII:
I am against this. There is a default pre channel, rest are just a backup option. Official channel is nZEDbPRE and that is it.

Wally73:
pre channel was dead this morning for a good hour already before i did put output of mine in it

[10:52:40] @Wally73 #nZEDbPRE output comes from my bot as the other bot left irc at [08:56:00] \* Quits: @nZEDbPreBot (CET), Turned output of mine on at [10:15:44] CET

niel:
@DariusIII I don't see the problem. This just adds one of the 'backups' to the default configuration so that users (most of whom know nothing about the alternatives) already have one in case of problems on the primary.
This kind of redundancy is the way we should go, IMO.
I can't add the bot I set up for IRCScraper to the primary because it would cause problems with double posting. Adding a second channel allows us to have a second or third channel for IRC-Scraper to retrieve data from.

kevinlekiller:
This is only if someone does not update their ircscraper config, the default in the config has 3 servers (which were added by DariusIII): 
https://github.com/nZEDb/nZEDb/blob/dev/nzedb/config/ircscraper_settings_example.php#L71

niel:
Good point @kevinlekiller, I'd forgotten that change after my forced 'sabbatical'.

PR #2180, opened 2016-06-16
Fixed SQL injection in Movie.php

pkostiuk:
Fixed a SQL injection in Movie.php. Feel free to reach out to me if you need working PoC.

### DIFF
--- a/nzedb/Movie.php
+++ b/nzedb/Movie.php
@@ -425,7 +425,7 @@ class Movie
 					$bbv .= '.';
 				}
 				if ($bb === 'imdb') {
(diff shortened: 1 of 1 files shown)


b3rs3rk:
I'm guessing the only thing that would block the attempt now is that $bbv can only be an integer based on the %d?

pkostiuk:
Yeah, I would normally use `$this->pdo->escapeString($bbv)` but I was just keeping to the style used elsewhere to mitigate SQLi for imdbID.

b3rs3rk:
Fair enough.  Let us know if you see anything else like that we should be concerned about.  I'll keep an eye out myself.

PR #2183, opened 2016-06-20
Remove 10-tvrage_titles.tsv and scripts used to pupulate that table.

DariusIII:
Addresses no specific issue.

- Remove 10-tvrage_titles.tsv
- Remove scripts used to populate that table, as they are not used anymore


b3rs3rk:
Was the table already removed from schema?

DariusIII:
The table itself does not exist in mysql_ddl.sql

niel:
Table was removed from schema so as not to create useless table for new users. The existing table hasn't had a patch to remove it because we were going to investigate using the existing data to populate the videos table, as I recall.

DariusIII:
That is useless now as we are using different method to populate those tables and tvrage_titles does not serve any purpose now.

b3rs3rk:
The table is trash, no use keeping it.

niel:
I was just giving some context, not arguing for keeping it. :-) Should we ever want the data it will still be there in history, so removing it isn't a problem.

Issue #2133, opened 2016-05-07
Can't update nZEDb when specifying a custom port / unix_socket


josh4trunks:
### Expected behaviour

`./zed update nzedb`
should update the database
### Actual behaviour

```
Unable to connect to host `localhost`.
```
### Steps to reproduce the behaviour

```
define('DB_HOST', 'localhost');
define('DB_PORT', '');
define('DB_SOCKET', '/mnt/mysql/mysql.sock');
```

josh4trunks:
There's at least 2 issues here.
In nZEDb, `app/config/bootstrap/connections.php` should have the below line added
`'unix_socket' => DB_SOCKET,`

In lithium, the following needs some work. It doesn't even use the $port they say to use.
https://github.com/UnionOfRAD/lithium/blob/master/data/source/database/adapter/MySql.php#L159-163
I'll work on getting lithium fixed and post here once it's done so we can pull in the fixes.

niel:
If you set the DB_HOST to '' it should work.
Oops, should have mentioned it is changed in dev, where we use 'socket' => DB_SOCKET,.

josh4trunks:
sounds good! actually looking at lithium's code, it does respect the passed socket, or even port. I'll get this fixed upstream.

niel:
you should be aware we're using the 1.1 branch NOT master.

josh4trunks:
sounds good, I'll make sure it gets fixed there.

josh4trunks:
Here's my PR for lithium's database handling.
https://github.com/UnionOfRAD/lithium/pull/1274

Once this gets merged, nZEDb won't need this logic since this would be handled in lithium. nZEDb can just pass the parameters directly
https://github.com/nZEDb/nZEDb/blob/dev/app/config/bootstrap/connections.php#L94-98

niel:
Thanks for the update.

josh4trunks:
@niel  UnionOfRAD/lithium ended up handling host/port/socket different than we expected.
They instead expect everything passed in the host variable like the below. See https://github.com/UnionOfRAD/lithium/blob/1.1/data/source/database/adapter/MySql.php#L173-190 

```
host=/path/to/socket
host=localhost
host=HOSTNAME:PORT
```

If needed I can submit a PR for the changes needed in https://github.com/nZEDb/nZEDb/blob/0.x/app/config/bootstrap/connections.php

[...]

PR #2189, opened 2016-06-24
Lithium Connections

josh4trunks:
Correctly passes parameters based on what lithium is expecting.
See https://github.com/UnionOfRAD/lithium/blob/1.1/data/source/database/adapter/MySql.php#L173-190

Fixes https://github.com/nZEDb/nZEDb/issues/2133 for all cases.

### DIFF
--- a/app/config/bootstrap/connections.php
+++ b/app/config/bootstrap/connections.php
@@ -91,7 +91,7 @@ if (file_exists($config)) {
 	}
 
 	if (isset($adapter)) {
-		$port = DB_SOCKET ?: (DB_PORT ? DB_HOST.':'.DB_PORT : DB_HOST);
+		$host = DB_SOCKET ?: (DB_PORT ? DB_HOST.':'.DB_PORT : DB_HOST);
 
 		Connections::add('default',
 			[
--- a/app/config/bootstrap/connections.php
+++ b/app/config/bootstrap/connections.php
@@ -91,20 +91,18 @@ if (file_exists($config)) {
 	}
 
 	if (isset($adapter)) {
-		$host = empty(DB_SOCKET) ? DB_HOST : DB_SOCKET;
+		$port = DB_SOCKET ?: (DB_PORT ? DB_HOST.':'.DB_PORT : DB_HOST);
 
 		Connections::add('default',
 			[
 				'type'       => 'database',
 				'adapter'    => $adapter,
 				'host'       => $host,
-				'port'       => DB_PORT,
 				'login'      => DB_USER,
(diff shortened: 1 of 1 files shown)


niel:
I don't understand this PR. What is it, that this is intended to achieve, that is currently missing?
Incidentally our coding policy doesn't permit complex ternaries, they are too complicated to read and often act differently to the intended action.

josh4trunks:
The current code doesn't pass the parameters correctly. I mentioned this before but I don't think it was noticed.
Lithium is expecting the host and port to be passed in the host variable like so

```
host=MYSQL_HOSTNAME:1234
```

Lithium doesn't use the socket or port variable.

I don't like the complex ternary either, just wanted to keep in line with what you had. I can change this to a normal if/else statement.

josh4trunks:
See this post, where I show what we should be passing to lithium
https://github.com/nZEDb/nZEDb/issues/2133#issuecomment-225236131

josh4trunks:
Ok, using standard if/else

niel:
I must be missing something. I looked at the Lithium code and as far as I can see it does NOT expect host:port in the host parameter.
https://github.com/UnionOfRAD/lithium/blob/1.1/data/source/database/adapter/MySql.php#L184 creates the dsn from both host and port settings.

Admittedly, the code before that is somewhat confusing as to its intent.

josh4trunks:
What you are looking at is the stuff after parsing. The only variable lithium pulls from configuration is host, which is parsed here https://github.com/UnionOfRAD/lithium/blob/1.1/net/HostString.php#L29

josh4trunks:
Here's the developers comment on it
https://github.com/UnionOfRAD/lithium/issues/1273#issuecomment-222825933

niel:
I was just reading the the Hostname class code. I think you're right, but I'll look at it a bit longer to be sure I'm getting it right.

josh4trunks:
Sounds good, take your time. Doesn't affect me until dev is merged in 0.x anyway.

niel:
Patch still misses one possibility. According to the Hostname class documentation it can accept ":<port>" and will automatically add the default host value.

niel:
BTW single level ternaries are acceptable. So the second level of that if structure can be a ternary.

josh4trunks:
I assumed nZEDb didn't accept that case so I didn't worry about it, but feel free to take over.

Ahh, I was thinking I could implement a ternary somewhere, but those are pretty new to me =]

josh4trunks:
sweet, looking forward to the next merge of dev

PR #2193, opened 2016-06-27
Replace usage of categoryid with categories in user_series and user_movies tables, respective classes and themes

DariusIII:
Addresses no specific issue..

- Replace categoryid in user_series table with categories_id
- Use categories_id in Releases class uSQL function
- Use categories_id in myshows.tpl in all of the themes


resources/db/schema/mysql-ddl.sql
```diff
@@ -1087,7 +1087,7 @@ CREATE TABLE user_movies (
   id          INT(16) UNSIGNED               NOT NULL AUTO_INCREMENT,
   user_id INT(16)                        NOT NULL,
   imdbid      MEDIUMINT(7) UNSIGNED ZEROFILL NULL,
-  categoryid  VARCHAR(64)                    NULL DEFAULT NULL,
+  categories_id  VARCHAR(64)                    NULL DEFAULT NULL,
```
niel: Really? Are you totally sure about this? I left it alone because it is a varchar and that implies it is being used totally differently to an integer. Everywhere else in the code I've seen a varchar used for category stuff, it was used to store a list of categories, not a single one. So you'd better be 101% sure of this before changing it. Same applies to SQL patch above.

b3rs3rk: I think it would be best represented as categories_ids?  Not sure if we have a standard for a column that would house multiple FKs at once to a table.

niel: We don't currently have a standard for that, but I'd prefer plain 'categories' as the _ids would make it too similar to categories_id and more prone to mistakes. Also, rows are usually singular, because they only contain a single value, so having it plural without the _id tells you it contains more than one value.
All of this is irrelevant if Darius is correct and the field contains a single value.

DariusIII: Tested and movies/series are added one per row, so categoryid/categories_id hold just one category in them, not a "list" of cats

niel: ok, you have my go ahead then.

DariusIII: No. Tested once more, series are added as : 5010|5030|5040 while movies are added as single category (but again that might be a bug). So it definitely should be named categories.

niel: ok, no great hurry to rename it.

DariusIII: Renamed categories_id into categories. Fix for adding movies same as tv shows will be done later, in another PR.

DariusIII:
Can someone review this, so we can merge it?

niel:
I already reviewed it, and gave my go ahead assuming it has been tested ;-)

Issue #2192, opened 2016-06-27
JSON output missing URL


ghost:
On 0.x branch:
Loading this URL: /api?apikey=&limit=1&o=json&offset=0&q=saving+hope&t=search

produces this JSON output:

``` json
[
  {
    "id": "10588016",
    "name": "(FnivatUbcr.F04R17R18-NalobqlFrraZlOnol] - \"FnivatUbcr.F04R17R18-NalobqlFrraZlOnol-YrgZrTb.par2\" yEnc",
    "searchname": "Saving.Hope.S04E17E18.1080p.WEB-DL.AAC2.0.H.264-NL",
    "totalpart": "137",
    "group_id": "1000011",
    "size": "4076798892",
    "postdate": "2016-04-26 16:14:34",
    "adddate": "2016-04-26 20:09:51",
    "updatetime": "2016-06-08 10:11:29",
    "guid": "1a00cbb6a846f9b1bb82577f8a287fdce627a3fe",
    "leftguid": "1",
    "fromname": "NLtranslated4U@gmail.com (Janeliz)",
    "completion": "100",
    "categories_id": "5040",
    "videos_id": "12",
    "tv_episodes_id": "407770",
    "imdbid": null,
    "xxxinfo_id": "0",
    "musicinfo_id": null,
    "consoleinfo_id": null,
    "gamesinfo_id": "0",
    "bookinfo_id": null,
    "anidbid": null,
    "predb_id": "0",
    "grabs": "2",
    "comments": "0",
    "passwordstatus": "0",
    "rarinnerfilecount": "3",
    "haspreview": "0",
    "nfostatus": "0",
    "jpgstatus": "0",
    "videostatus": "0",
    "audiostatus": "0",
    "dehashstatus": "0",
    "reqidstatus": "0",
    "nzb_guid": "Ã…ÃœÂ›Ã¦BÃ'Â®\nÃžÂ'-HsngÂ¿",
    "nzbstatus": "1",
    "iscategorized": "1",
    "isrenamed": "1",
    "ishashed": "0",
    "isrequestid": "0",
    "proc_pp": "0",
    "proc_sorter": "0",
    "proc_par2": "1",
    "proc_nfo": "0",
    "proc_files": "1",
    "category_name": "TV > HD",
    "0": "0",
    "category_ids": "10",
    "failed": null,
    "group_name": "alt.binaries.boneless",
    "nfoid": null,
    "reid": null,
    "categoryparentid": "5000",
[...]

PR #2194, opened 2016-06-27
Improve JSON Response from API/RSS by basing it off of XML Response

b3rs3rk:
Addresses issue #2192 JSON output missing URL.

- Remove JSON_Response Class.
- Add function to convert XML_Response return to Assoc Array.
- When JSON output is requested, the XML Response is converted to an associative array whilst preserving the XML name spaces and then encoded as JSON.


nzedb/http/Capabilities.php
```diff
@@ -69,30 +70,28 @@ public function output($data, $params, $xml = true, $type = '')
...
 
+		$options = [
+			'Parameters' => $params,
+			'Data'       => $data,
+			'Server'     => $this->getForMenu(),
+			'Type'       => $type
+		];
+
+		$response = (new XML_Response($options))->returnXML();
+
```
niel, line 81: Couldn't this go inside the IF. No point spending the cycles creating this, only to overwrite it if JSON is wanted.

nzedb/utility/Text.php
```diff
@@ -211,4 +211,90 @@ public static function convertRomanToInt($string)
 		}
 		return $e;
 	}
+
```
niel: Not sure this belongs in Text, but I can't think of anywhere more appropriate ;-)

b3rs3rk: Huh, I only added the extra line there though.  I moved it to Misc where we have another object2array function.

nzedb/http/Capabilities.php
b3rs3rk, line 81: The JSON response is now dependent upon the XMLWriter response, so it is needed.  This way we only need to manage one class for arranging the data.  We take a small performance hit because we then have to load the XML data and convert it to JSON if JSON was requested.

DariusIII:
Using this in nntmux with no issues so far. As far as i am concerned, this is good to go.

ghost:
Yeah @niel merge this in august please 👍

Issue #2241, opened 2016-08-25
Just some random stuff


cytec:
While i was installing and setting up nZEDb i've found some things not really covered in the install docs as well as some minor things that don't seem to work in templates.
1. If you'll want to show more than x releases on /movies page or /movies/?imdb=123456 you'll need to set a higher value for mysql's `group_concat_max_len` the default value is `1024` which is in most cases something around 9-14 Releases. I've increased it to `100000` which fixed those issues for me. (can provide Screenshots if needed) maybe this should be added to FAQ or something.
2. on the `moviesfullview.tpl` template the buttons for `add to cart` and `sabnzdb` are not working. Simply copy&paste them from `browse.tpl` fixed that. (Alpha theme) see: https://github.com/cytec/nZEDb/commit/0313f3aa574635051a8bb7d0cf4eb64aa332b660
3. profile view by name (for admins) isn't working as there is no value set for `$userID` so it'll show always the admin profile. ID is working as expected. see: https://github.com/cytec/nZEDb/commit/bc5f29618ae869a0575c970db3a266631f2dad74
4. AniDB will never get infos from remote as the query checking if an remote update is needed for `anidb_info` always returns `false` if the show doesn't exist in database. see: https://github.com/cytec/nZEDb/commit/470536237aafaad61aed15f9b936e22635b5eb37 even that doesn't handle both but just updates the entry ONCE from remote when it is created.

kevinlekiller:
Thanks for the fixes.

About the `group_concat_max_len`, in the documentation I wrote, it is included, for example: https://github.com/nZEDb/nZEDb/blob/0.x/INSTALL.txt#L44 
https://github.com/nZEDb/nZEDb_Misc/blob/master/Guides/Installation/Ubuntu/Guide.md#step-7-configuring-mysql

The other guides / docs are community provided, it's possible they forgot to include that information.

cytec:
@kevinlekiller nevermind then... seems like i just didn't saw it here: https://github.com/nZEDb/nZEDb_Misc/blob/master/Guides/Installation/Ubuntu/Guide.md#step-7-configuring-mysql

kevinlekiller:
No worries, it's a long doc, easy to miss to something.

PR #2244, opened 2016-08-25
Update isTV regex to match full season releases

cytec:
Addresses issue #2241.

This contains only the regex part


kevinlekiller:
Looks good to me if anyone wants to merge. Tested it in a regex tester.

PR #2245, opened 2016-08-25
Template fixes

cytec:
Addresses issue #2241.
only the template stuff

- fix profile page by name for admin/mods
- fix library includes 
- fix forum view
- fix fullmoviesview buttons for releases
- show movie covers in viewnzb

### DIFF
--- a/www/themes/Alpha/templates/viewmoviefull.tpl
+++ b/www/themes/Alpha/templates/viewmoviefull.tpl
@@ -116,6 +116,7 @@
 			{assign var="mhaspreview" value=","|explode:$result.grp_haspreview}
 			{assign var="mcatname" value=","|explode:$result.grp_release_catname}
 			{foreach $msplits as $m}
+				{if $m@index < $mname|@count}
 				<tr class="{cycle values=",alt"} filter"data-name="{$mname[$m@index]|escape:"htmlall"|replace:".":" "|lower}" id="guid{$mguid[$m@index]}">
 					<td class="check"><input id="chk{$mguid[$m@index]|substr:0:7}" type="checkbox" class="nzb_check" value="{$mguid[$m@index]}" /></td>
 					<td class="item">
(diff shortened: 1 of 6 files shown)


www/themes/Alpha/templates/viewmoviefull.tpl
```diff
@@ -141,23 +142,16 @@
...
+							<a title="Download Nzb" href="{$smarty.const.WWW_TOP}/getnzb/{$mguid[$m@index]}"></a>
+						</div>
+						{if $sabintegrated}
+							<div class="icon icon_sab" title="Send to my Queue"></div>
+						{/if}
+						<div class="icon icon_cart" title="Add to Cart"></div>
 					</td>
 				</tr>
+				{/if}
 			{/foreach}
```
DariusIII, line 155: I don't see a reason for these changes.  Looking at the code, these will not work. And you have removed font-awesome icons on the way.

cytec: the reason for those changes: it didn't work and didn't show the icons correctly in the first place especially the button part...  

i used the exact same code you use in `browse.tpl` to show the icons see: https://github.com/nZEDb/nZEDb/blob/dev/www/themes/Alpha/templates/browse.tpl#L221-L228 

the `{if $m@index < $mname|@count}` part is just to make sure there are only releases shown which do have a name... depending on your `group_concat_max_len` there might be some rows that don't have a release name otherwise

EDIT: here are 2 images before and after the above changes... the fa icons didn't work anyway as the css wasn't included corretly 

Before the change
after

DariusIII: All of this is because somehow my changes to alpha got reverted,

cytec: didn't know that.. then feel free to ignore this PR

PR #2251, opened 2016-08-29
Add by poster categorization

DariusIII:
Addresses no issue.

- Add categorization of releases by poster
- Change determineCategory function to also use poster.


kevinlekiller:
Can you add support for the DB regexes ?

Like this : https://github.com/nZEDb/nZEDb/blob/0.x/resources/db/schema/data/10-category_regexes.tsv

PR #2266, opened 2016-09-14
Refactor yydecode to use std I/O instead of files, improves performance.

kevinlekiller:
Addresses no issue.

- Pass yEncoded data to yydecode through stdin
- Read decoded data from yydecode from stdout
- Pass yydecode -b option only if $ignore is true

Test: https://gist.github.com/kevinlekiller/fcec2902b41fd70eb57dc15d31af5bdf


niel (approved):
Will merge in later today.

niel:
Idea was merged into the adapter, not an actual code merge though.

Issue #2277, opened 2016-10-04
Can't delete releases based on category id


ghost:
php delete_releases.php tegories_id=equals=1000

Output:
Error: Invalid argument supplied: categories_id=equals=1000

niel:
1000 is a category parent, there shouldn't be any releases under it directly. They will be under a more specific category or <catname>_OTHER

ghost:
It isn't the categoryID. It didnt work on any of them.

ghost:
Same problem here

ghost:
change line 1158 to  
 case 'categories_id':
 in nzedb/ReleaseRemover.php

ghost:
That worked. Created a PR for it.

niel:
Damnit. The delete_releases.php shouldn't have been changed. I'll revert that change, please go back to using categoriesid.

PR #2289, opened 2016-10-15
Changed categoryid to categories_id

ghost:
Thanks to @slug1982 who found the wrong variable for this issue #2277


niel:
Can you explain what problem this fixes? Is it connected to use of .../misc/testing/Release/delete_releases.php, for example.

niel:
Fixed the help text in delete_releases.php script to fix this.

PR #2293, opened 2016-10-16
remove rogue semicolon

MikeOsborn:
This extra semicolon caused the database to be partially created causing further issues with the site after the install process was complete.


DariusIII:
I have to write 2 things:
- All PR's need to go against the dev branch
- This has already been addressed in commit 3ba52fbeddfdde1ae534e8e235fefaef207ed343 in dev branch.

PR #2305, opened 2016-10-29
Replace php-gd with php-imagick in ReleaseImage class, should fix segfaults in PPA

DariusIII:
Addresses issue #.
## 
##


niel (approved):
I haven't been able to test, but it looks good to me.

niel (approved):

PR #2350, opened 2017-01-05
Dev mgr

DariusIII:
Addresses issue #.

No issue is addressed

- Update handling of releases creation by using updated method.


misc/update/nix/multiprocessing/.do_not_run/switch.php
```diff
@@ -230,7 +232,7 @@
 			collectionCheck($pdo, $options[2]);
 
 			// Create releases.
-			processReleases($pdo, new ProcessReleases(['Settings' => $pdo]), $options[2]);
+			processReleases($pdo, new ProcessReleases(['Settings' => $pdo]), new ReleasesMultiGroup(), $options[2]);
```
niel: Still using $pdo

nzedb/Binaries.php
```diff
@@ -615,21 +616,32 @@ public function scan($groupMySQL, $first, $last, $type = 'update', $missingParts
...
 		$notYEnc = $headersBlackListed = 0;
 
 		$partsQuery = $partsCheck = sprintf('INSERT IGNORE INTO %s (binaryid, number, messageid, partnumber, size) VALUES ', $tableNames['pname']);
+		$mgrPartsQuery = $mgrPartsCheck = sprintf('INSERT IGNORE INTO mgr_parts (binaryid, number, messageid, partnumber, size) VALUES ');
 
 		$this->_pdo->beginTransaction();
 		// Loop articles, figure out files/parts.
 		foreach ($headers as $header) {
 
+			ReleasesMultiGroup::isMultiGroup($header['From']) ? $multiGroup = true : $multiGroup = false;
```
niel: Style: use ternary to set a variable then use the variable

nzedb/Binaries.php
```diff
@@ -724,27 +748,33 @@ public function scan($groupMySQL, $first, $last, $type = 'update', $missingParts
 					// Get the current unixtime from PHP.
 					$now = time();
 
+					$xref = ($multiGroup === true ? sprintf('xref = CONCAT(xref, "\\n"%s ),', $this->_pdo->escapeString(substr($header['Xref'], 2, 255))) : '');
```
niel: This still feels kinda dirty to me. XREF is data that comes from the usenet server, we shouldn't change it, any changes needed should go somewhere else so that the original value is always correct.

nzedb/Binaries.php
niel: looks like it's over complicated.
why not just return the value directly to $multiGroup? i.e. 
$multigroup = ReleasesMultiGroup::isMultiGroup($header['From']);

niel: In other words, the intent here is not at all clear.

nzedb/Binaries.php
```diff
@@ -724,27 +748,33 @@ public function scan($groupMySQL, $first, $last, $type = 'update', $missingParts
...
 					$collectionID = $this->_pdo->queryInsert(
 						sprintf("
 							INSERT INTO %s (subject, fromname, date, xref, group_id,
 								totalfiles, collectionhash, dateadded)
 							VALUES (%s, %s, FROM_UNIXTIME(%s), %s, %d, %d, '%s', NOW())
-							ON DUPLICATE KEY UPDATE dateadded = NOW(), noise = '%s'",
-							$tableNames['cname'],
+							ON DUPLICATE KEY UPDATE %s dateadded = NOW(), noise = '%s'",
+							($multiGroup === true ? 'mgr_collections' : $tableNames['cname']),
```
niel: Style: Ternaries should always be on their own line, never part of a more complex statement.

nzedb/Binaries.php
```diff
@@ -760,8 +790,8 @@ public function scan($groupMySQL, $first, $last, $type = 'update', $missingParts
 						INSERT INTO %s (binaryhash, name, collection_id, totalparts, currentparts, filenumber, partsize)
 						VALUES (UNHEX('%s'), %s, %d, %d, 1, %d, %d)
 						ON DUPLICATE KEY UPDATE currentparts = currentparts + 1, partsize = partsize + %d",
-						$tableNames['bname'],
-						md5($matches[1] . $header['From'] . $groupMySQL['id']),
+						($multiGroup === true ? 'mgr_binaries' : $tableNames['bname']),
```
niel: See line 758 above.

nzedb/Binaries.php
```diff
@@ -809,7 +849,8 @@ public function scan($groupMySQL, $first, $last, $type = 'update', $missingParts
 		// End of processing headers.
 		$timeCleaning = number_format($startUpdate - $startCleaning, 2);
 
-		$binariesQuery = $binariesCheck = sprintf('INSERT INTO %s (id, partsize, currentparts) VALUES ', $tableNames['bname']);
+		$bTable = ($multiGroup === true ? 'mgr_binaries' : $tableNames['bname']);
```
niel: Parentheses not needed.

nzedb/Binaries.php
```diff
@@ -1308,6 +1398,22 @@ private function addMissingParts($numbers, $tableName, $groupID)
 	}
 
 	/**
+	 * Add article numbers from missing headers to DB.
+	 *
+	 * @param array  $numbers   The article numbers of the missing headers.
+	 *
+	 * @return bool
+	 */
+	private function addMissingMGRParts($numbers)
```
niel: add 'array' to the definition if it MUST be an array. This causes an error if non-arrays are passed making debugging easier

nzedb/Binaries.php
```diff
@@ -1326,6 +1432,22 @@ private function removeRepairedParts($numbers, $tableName, $groupID)
 	}
 
 	/**
+	 * Clean up mgr_missed_parts table.
+	 *
+	 * @param array  $numbers   The article numbers.
+	 *
+	 * @return void
+	 */
+	private function removeRepairedMGRParts($numbers)
```
niel: Same as above

nzedb/Binaries.php
DariusIII: Intent is to set a flag for MGR so second set of tables is used for MG releases.

nzedb/Binaries.php
DariusIII: We actually never use xref value for anything up until recently. Now we use Xref to add xposted groups to nzb. As collections are deleted upon release creation, there is no need for "original" xref

niel: You missed the point. We're not using the 'XREF' value, this uses a corrupted version of it. We shouldn't be doing that. Instead we should store that data somewhere else, like a new table or column.

DariusIII: Why would you add another column or table for something that is deleted after release has been created? This is used only for MGR, as those are posted to 8 or 9 groups with only one group in xref. This is done solely for NZB creation, so we have proper groups listed there.

niel: if it is done soley for NZB creation, AND it's done at the same time, then may as well create the array/list in RAM. If on the other hand it is done ad hoc (whenever a new group is discovered during header processing) then longer term storage is needed: hence table or column

b3rs3rk:
@niel - this is ready for review and pending agreement merge

DariusIII:
This is definitely ready to be merged.

misc/update/nix/tmux/monitor.php
```diff
@@ -217,7 +217,7 @@
 						$stamp = 'UNIX_TIMESTAMP(MIN(dateadded))';
 
 						switch (true) {
-							case strpos($tbl, 'collections_') !== false:
+							case strpos($tbl, 'collections') !== false:
```
niel, line 220: Is this intended to match "collections_*" and "multigroup_collections"? It's possible this would come back and bite us if we ever added another table with 'collections' in the name (same applies to the following 'binaries' and 'parts' checks too). It would be safer to use a match at pos 0 and be more specific with the names. i.e.
case strpos($tbl, 'collections') === 0 || strpos($tbl, 'multigroup_collections') === 0:
assuming the intent is to match both. This way the name has to start at column 0 and will avoid any we add later that don't

misc/update/nix/tmux/monitor.php
```diff
@@ -227,17 +227,17 @@
 									$age = $added['dateadded'];
 								}
 								break;
-							case strpos($tbl, 'binaries_') !== false:
+							case strpos($tbl, 'binaries') !== false:
```
niel, line 230: See above.

misc/update/nix/tmux/monitor.php
```diff
@@ -227,17 +227,17 @@
...
 								$runVar['counts']['now']['binaries_table'] +=
 									getTableRowCount($psTableRowCount, $tbl);
 								break;
 							// This case must come before the 'parts_' one.
-							case strpos($tbl, 'missed_parts_') !== false:
+							case strpos($tbl, 'missed_parts') !== false:
```
niel, line 235: See above.

misc/update/nix/tmux/monitor.php
```diff
@@ -227,17 +227,17 @@
...
 								$runVar['counts']['now']['missed_parts_table'] +=
 									getTableRowCount($psTableRowCount, $tbl);
 
 								break;
-							case strpos($tbl, 'parts_') !== false:
+							case strpos($tbl, 'parts') !== false:
```
niel, line 240: See above.

www/admin/posters-edit.php
```diff
@@ -0,0 +1,56 @@
...
+		if (!empty($_GET['id'])) {
+			$page->title = "MultiGroup Poster Edit";
+			// Note: explicitly setting default stuff below, which could be shortened to:
+			// $entry = MultigroupPosters::find($_GET['id']);
+			$entry = MultigroupPosters::find('first',
+				[
+					'conditions' => ['id' => $_GET['id']],
+					'fields' => ['id', 'poster']
+				]);
+			$poster = [
```
niel: This can be simplified to just $poster = MultigroupPosters::find('first' ...
and let have the template use the object directly, has the added bonus of being faster too.

niel:
I added some comments if they, and the above merge conflict, are addressed I'm good to merge it into dev next week.

nzedb/Binaries.php
```diff
@@ -637,14 +637,20 @@ public function scan($groupMySQL, $first, $last, $type = 'update', $missingParts
 		// If there was an error, try to reconnect.
 		if ($this->_nntp->isError($headers)) {
 
+			if ($this->first == $this->last) {
```
niel, line 640: I didn't have a problem with the ternary itself, just having it in the sprintf. This if/else is good to be a ternary.
Ternaries should have a line of their own and assign a variable, according to our code standard.

b3rs3rk: Doesn't really matter.  The whole point of doing the ternary in the sprintf (I believe) in the first place was to save memory as part of the optimizations @kevinlekiller did when he tried to optimize binary updates as much as possible.  So if the requirement is to set it to a var for sprintf, it defeats the purpose of using the ternary which is much more annoying to look at than an if/then/else.

niel: I don't know if that was the intent with that ternary or not, but it makes reading the code more complicated in general, which is why we have the coding standard. 
IF the intent is for micro-optimising, that has to be considered separately and might override normal coding standards. That's a pretty big IF though.

misc/update/nix/tmux/monitor.php
b3rs3rk: Wouldn't it just be better to make the SQL query be something more descript than 'SHOW TABLE STATUS' and not worry about the specificity of the logic gates in that switch?  Something more like what we do with the TPG patch changes.

Honestly, the more I think about it, this whole procedure could likely be done with an SP and reduce the PHP code requirement altogether.

PR #2351, opened 2017-01-07
Dev db

niel:
- Binaries table: collection_id changed to collections_id
- Collections and missed_parts tables: group_id changed to groups_id
- Parts table: binaryid changed to binaries_id
- DDL, patches, and code changed to reflect above items.

Note there are still some 'group_id' strings deliberately left in the code. This is because I'm not certain that they are references to a table column and might just be an arbitrary identifier name. Let me know if you see any that should be changed.


DariusIII:
As far as i am concerned, this PR is good to go.

DariusIII (approved):

PR #2365, opened 2017-01-23
Update resetdb.php

krabsheva:
Newer table "releases_groups" wasn't added to script. Caused subsequent runs of release update scripts to silently exit leaving "duplicate PRIMARY key" errors in logs. Perhaps also missing from other scripts but don't have the knowledge of code base to hunt them all down.

### DIFF
--- a/misc/testing/DB/resetdb.php
+++ b/misc/testing/DB/resetdb.php
@@ -39,7 +39,8 @@ $arr = [
 		"videos", "tv_episodes", "tv_info", "release_nfos", "release_comments", 'sharing', 'sharing_sites',
 		"users_releases", "user_movies", "user_series", "movieinfo", "musicinfo", "release_files",
 		"audio_data", "release_subtitles", "video_data", "releaseextrafull", "parts",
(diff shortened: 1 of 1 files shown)


kevinlekiller:
You could get a list of tables not in that array, but I don't know if they all need to be truncated:

    arr=$(curl -sL https://github.com/nZEDb/nZEDb/raw/0.x/misc/testing/DB/resetdb.php | pcregrep -M "$arr = \[(\s|.)*?\];" | sed "s/\(^\$arr = \[\|\];\|\,\)//g" | xargs)
    for table in $(curl -sL https://github.com/nZEDb/nZEDb/raw/0.x/resources/db/schema/mysql-ddl.sql | grep -oi "create table .*(" | sed "s/\(^create table\s*\|\s*($\)//gI"); do if [[ ! $arr =~ "$table" ]]; then echo $table; fi; done

niel:
Not all table need to be truncated. PreDb is a good example of one that most would want kept.

@krabsheva Please create separate branches for future PRs. This one had the same name as the previous one, which causes errors for merging.

PR #2375, opened 2017-02-02
Remove reference to readthedoc install guide.

kaibosh:
Point to wiki so that future install doc updates don't require PR.

Addresses issue #.

-
-
-


kaibosh:
Note: PR #2376 also has this reference removed.

PR #2379, opened 2017-02-07
Update sequential scripts for mandatory TPG use

kaibosh:
Addresses issues reported in IRC.

- Update sequential scripts to use TPG compatible update_releases scripts.
- Change simple-expanded.sh to use PHP rather than python upd_release scripts.
-

### DIFF
--- a/misc/update/nix/screen/sequential/README.txt
+++ b/misc/update/nix/screen/sequential/README.txt
@@ -4,4 +4,8 @@ Edit the path to your nZEDb location.
 
 start like this:
 
-screen sh start.sh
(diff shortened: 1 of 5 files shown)


Wally73:
line 45 $PHP $THREADED_PATH/releases.php should be
$PHP ${THREADED_PATH}/releases.php

line 48 $PHP $NZEDB_PATH/postprocess.php all true should be 
$PHP ${NZEDB_PATH}/postprocess.php all true

kaibosh:
Done, thanks Wally73

PR #2406, opened 2017-03-05
Fix `zed update nzedb`

egyptianbman:
Updating was failing prematurely because `Update::nzedb()` expects `Update::git()` to return the output of the `Update::git->pull()`. This commit adjusts `Update::git()` to not only output the response from `Update::git->pull()`, but also to return the trimmed value for evaluation in `Update::nzedb()`, allowing the update to continue.

### DIFF
--- a/app/extensions/command/Update.php
+++ b/app/extensions/command/Update.php
@@ -111,7 +111,10 @@ class Update extends \app\extensions\console\Command
 			return;
 		}
 
-		$this->out($this->git->pull());
+		$output = $this->git->pull();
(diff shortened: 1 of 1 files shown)


niel:
Thanks for the fix. I don't know how I could have missed this. ;-)
Please see https://github.com/nZEDb/nZEDb/wiki/Contributing-Code-For-Beginners-to-Git(hub) for our policies on PRs. Mainly, note to use a branch for PRs. I'll merge this in as is this time.

PR #2409, opened 2017-03-08
Dev refactor tpg

niel:


niel:
Oops PRd against 0.x, this is merged into dev.

PR #2412, opened 2017-03-09
Fix path to config.php

jangrewe:
My backfill pane shows an error after updating to 0.7 because info.py still has the old location of config.php set.

### DIFF
--- a/misc/update/python/lib/info.py
+++ b/misc/update/python/lib/info.py
@@ -50,7 +50,7 @@ def disconnect(cur, con):
 
 
 def readConfig():
-    Configfile = pathname+"/../../../nzedb/config/config.php"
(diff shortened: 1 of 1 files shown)


kaibosh:
Try switching the tmux backfill mode to "All" as this will use the php backfill scripts. The backfill-groups mode is using the non-supported python scripts and will probably be removed in the near future.

jangrewe:
Alright, that will fix it. Merging this PR for anybody who still uses the Python scripts probably won't hurt, though ;-)

niel:
We do not accept PRs against stable branch. Please look at this https://github.com/nZEDb/nZEDb/wiki/Contributing-Code-For-Beginners-to-Git(hub) for future reference. In the mean time I'll do this manually in dev branch.

Issue #2437, opened 2017-04-14
Joins without indexes and overall high CPU usage


ghost:
I've been having issues with high MySQL CPU usage lately. My avg. CPU usage tripled from 25% 10 months ago to 80+% today.

I have enabled the slow query log and logged joins without indexes. The log file filled at breakneck speed with these two queries:

```sql
SELECT COUNT(z.id) AS count FROM (SELECT r.id FROM releases r
			LEFT OUTER JOIN videos v ON r.videos_id = v.id AND v.type = 0
			LEFT OUTER JOIN tv_info tvi ON v.id = tvi.videos_id
			LEFT OUTER JOIN tv_episodes tve ON r.tv_episodes_id = tve.id
			LEFT JOIN categories c ON c.id = r.categories_id
			LEFT JOIN categories cp ON cp.id = c.parentid
			LEFT JOIN groups ON groups.id = r.groups_id
			LEFT OUTER JOIN video_data re ON re.releases_id = r.id
			LEFT OUTER JOIN release_nfos rn ON rn.releases_id = r.id

			WHERE r.categories_id BETWEEN 5000 AND 5999
			AND r.nzbstatus = 1
			AND r.passwordstatus <= 0
			AND r.tv_episodes_id IN (485589)   AND r.categories_id IN (5010, 5020, 5030, 5040, 5045, 5060, 5070, 5080, 5999, 5070, 5080, 5020, 5040, 5999, 5030, 5060, 5045, 5010)  AND r.postdate > NOW() - INTERVAL 2500 DAY  LIMIT 125000) z;


SELECT users.*, user_roles.apirequests, user_roles.downloadrequests, NOW() AS now
				FROM users
				INNER JOIN user_roles ON user_roles.id = users.role
				WHERE LOWER(users.rsstoken) =
				LOWER('775e155c780ed5af9119f797f814c714') LIMIT 1;
```

When doing `SHOW FULL PROCESSLIST` I can see a LOT of processes of the last query. It's a simple query but it takes 700ms+ to 1,5s. I guess it should easily complete within 100-150ms.

The first query I see a lot too.

MySQL Tuner shows: `[!!] Joins performed without indexes: 968652` and `[!!] Temporary tables created on disk: 59% (29M on disk / 49M total)`

Shouldn't there be indexes for these queries?

ghost:
[...]

PR #2438, opened 2017-04-14
Speed up getByRssToken function

ghost:
Addresses issue #2437 

- Added index to users table
- Removed MySQL LOWER() function from getByRssToken()

### DIFF
--- a/nzedb/Users.php
+++ b/nzedb/Users.php
@@ -537,8 +537,7 @@ class Users
 				SELECT users.*, user_roles.apirequests, user_roles.downloadrequests, NOW() AS now
 				FROM users
 				INNER JOIN user_roles ON user_roles.id = users.role
-				WHERE LOWER(users.rsstoken) =
-				LOWER(%s)",
+				WHERE users.rsstoken = %s",
 				$this->pdo->escapeString($rssToken)
 			)
 		);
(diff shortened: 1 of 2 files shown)


b3rs3rk:
Need to add the index to the DDL file as well @Fossil01

ghost:
Will do when im back home.

niel:
Also need to address the underlying issue, make sure those values are already lower-cased so the check is not needed in the SQL. I can help with that part of it.

ghost:
I think the values always are and always were lowercase. It's an API key, never seen caps being used.

niel:
It still needs to be checked before going into the table to be certain it is. Which is how it should have been done originally as it is less processing required to sanitize it on the PHP side.

niel:
I have some time now, I'll merge in the PR and add the extra stuff needed.

niel:
Arrrgh! you forgot to branch.

niel:
and hardcoded the patch number!!

ghost:
Committed from dev to dev heh. I was in a hurry 😝

ghost:
Delete the pr if u will. Can fix tonight.

niel:
I'm working through it already. Will finish it off

ghost:
Hey @niel I have found more of these LOWER() queries which are not needed and can be replaced with strtolower.

https://github.com/nZEDb/nZEDb/blob/8e32d7213a9731bfa601ec731ff05120af756f91/misc/testing/Various/sabnzbd_control.php#L23

https://github.com/nZEDb/nZEDb/blob/2633d35edacd79bfcfef663924a8f29694a5da99/nzedb/Users.php#L435

https://github.com/nZEDb/nZEDb/blob/2633d35edacd79bfcfef663924a8f29694a5da99/nzedb/Users.php#L448

PR #2445, opened 2017-05-02
Reworked all Anime release naming regexes

nightah:
Complete cleanup of all Anime based regexes:
- Removed all duplicate anime regexes
- Consolidated anime regexes to 3 "generic" regexes
- Realigned ordinals to prefer "generic" regexes

Regex tests below (obviously global and multi line are only turned on for the purpose of the test):
33: https://regex101.com/r/Yus6Jo/5
34: https://regex101.com/r/3sysKA/7
35: https://regex101.com/r/mvugyg/2


niel:
I would like to know your reasoning for changing the preference to generic regexes? All of our regexes are aimed at specific ones first, gradually getting more generic, and is by design.

nightah:
When I say generic, I mean the genericised regexes that I have written specific to the anime release groups, I hope that clears up any initial confusion.

Having said that, I've been heavily monitoring the anime release groups and almost everything I have seen up until 01/01/2017 (and even beyond that date) fall within the 3 new regexes that I have included. The others that have been removed obviously fell within the 3 that I wrote, and the rest that did not still exist. However, the ones that still exist have also been set lower ordinals because frankly I haven't seen them be triggered at all, but I don't think that's a reason to remove them, at least for now.

I guess part of the problem here is that anime isn't heavily regulated like scene releases with specific naming conventions, add that with the inconsistencies of some of the posters and you have 10+ regexes based on each use case which may or may not be relevant anymore.

nightah:
It's also worth noting during my testing practically all of the anime releases that end up falling outside of the revised regexes are due to the fact that the search name is more than 255 characters. Therefore the name is being cut off and is not caught by the respective regexes.

niel:
We're exploring extending the size of the name field. Main problem is how it affects indexes (limited length)

PR #2458, opened 2017-05-30
Comment out use of "placeholder" tmux session. The error output it wa…

kaibosh:
…s suppressing no longer appears.
Note: Commenting out for now. Will remove later should no problems occur.

Add warning message regarding usable nzedb modes if using tmux v2.3 or higher.
-
-
-


kaibosh:
Tested this change using all 3 nzedb modes with tmux v1.8 and 2.0.

As it's almost end of month we could merge this later for the July release and give it more testing time in dev.

PR #2460, opened 2017-05-31
Fix bad formating in 10-release_naming_regexes.tsv

nightah:
Currently if you create a new database with the existing tsv, records up until 1148 are inserted. The subsequent records are inserted into 1148's description.

- Fix bad formating in 10-release_naming_regexes.tsv


nightah:
@niel this is what I was discussing with you on IRC.

Let me know if you also want a patch alongside this, to reinsert 1148-1156 for anybody that created a new db within the period that the tsv was broken.

Issue #2472, opened 2017-06-12
Ahem - Adult Releases not renamed ?


PsychoSid:
###Expected behaviour
Adult releases not being renamed correctly ?

###Actual behaviour
Should be renamed

###Steps to reproduce the behaviour
Run updates etc

Not that I am especially bothered (honest !) but just running through things on the site since the update and noticed that at of all things the Adult stuff seems not to be named/renamed correctly all others seem fine:-
$ ./zed version all
Looking up Git tag version(s)
Hash: 84a81108a9112f0b018215e6696f02bf88f09e3c
XML version: 0.7.3.1
Git version: 0.7.3.1
Looking up SQL patch version(s)
XML version: 482
 DB version: 482

< wckdpctrs my brthrs grlfrnd fllhd >-[01/97] - "wckdpctrs my brthrs grlfrnd fllhd par2"  for example.

Bart39:
That's not going to match anything, not a scene release or even anything to work out quality etc

PsychoSid:
Yeah - dunno to be fair. Just know I got a lot of human readable releases prior to the latest update.

Thanks

sspilleman:
Not related to (the last version of) nzedb I guess. Looks like its a new thing to post stuff in that manner.... See:

https://www.nzbindex.nl/search/?q=wckdpctrs&age=&max=25&minage=&sort=agedesc&minsize=&maxsize=&dq=&poster=&hidespam=0&hidespam=1&more=0

sspilleman:
Please ignore my previous comment. Although its true, a lot of names are mangled... In this it *should* have renamed it... Its in PreDB.

bgidi:
It's because the posters at a.b.e now use a new format.

It's no longer:

    [278997]-[FULL]-[#a.b.erotica]-[ chi-the.walking.dead.xxx ]-[06/51] - "chi-the.walking.dead.xxx-s.mp4" yEnc

It's now: 

    < wckdpctrs.my.brthrs.grlfrnd.hd >-[01/47] - "wckdpctrs.my.brthrs.grlfrnd.hd.par2" yEnc

or

    [ ntgrl.17.06.14.y.4k ]-[01/94] - "ntgrl.17.06.14.y.4k-Sample.mp4" yEnc 

The ID are now missing from the subject but filenames have corresponding entries in the PreDB.

I think the collection regex should be something like:

[...]

PR #2484, opened 2017-06-22
Refactor XXX releases fetching support

DariusIII:
Addresses issue #.
No XXX info pulled from some of the sources, should fix issue #2472 
- Refactor XXX releases support
- Fix matching to all of the sources


app/extensions/helper/helpers.php
```diff
@@ -0,0 +1,44 @@
+<?php
```
niel, line 1: This is in the wrong place. lithium extensions are for extending a base lithium class.
Put this in \nzedb\utilities\misc.php for now

composer.json
```diff
@@ -5,6 +5,9 @@
     }
```
niel, line 5: This won't be needed after moving the function to  \nzedb\utilities\misc.php

app/extensions/helper/helpers.php
DariusIII: I have used laravel logic, we do have getUrl in misc and i did not want to mix the two.

niel: so rename yours?
 ;-)

DariusIII: Mine is getRawHtml :)

niel:
Closed for Darius to create a new PR.

PR #2487, opened 2017-06-26
Refactor XXX to properly process XXX releases

DariusIII:
Addresses issue #.

Fixes issue #2472

- Move XXX related classes to nzedb/processing/adult
- Update and fix ADE, ADM, AEBN, HOTMOVIES, POPPORN classes
- Use guzzlehttp to fetch raw HTML
- Add getRawHtml function to Misc class


composer.json
```diff
@@ -131,7 +131,8 @@
     "bower-asset/select2-4.0.x": "~4.0.2",
     "bower-asset/slimScroll-1.3.x": "~1.3.7",
     "bower-asset/tinymce-builded": "^4.4",
-    "zeebinz/rarinfo": "~2.5.0"
+    "zeebinz/rarinfo": "~2.5.0",
+    "guzzlehttp/guzzle": "^6.3"
```
niel: Please put this in the list alphabetically, so it is easier to find when looking through the list.

DariusIII: FYI, if i add that manually, i need to run composer update, which i don't want to do, because it will update rest of the libs. 

So i used

 `composer require guzzlehttp/guzzle. `

Next time you plan updating composer libraries, just move that line to be sorted alphabetically.

niel: I wasn't asking you to add it manually, just move it manually exactly as you asked me to do later. I know Composer always adds at the end, which is a PITA.

PR #2498, opened 2017-08-02
Implemented TMDB support for Upcoming Movies section in Alpha Template

zetas:
RT is broken, this adds TMDB support. Still needs to be updated via the same old update script every 12 hours.


joeynwz (approved):
Good alternative to RT

PR #2602, opened 2018-10-24
FIX anidb_info -> categories

JReming85:
I don't know if anidb changed this from categories, or just removed it altogether. What I do know is it leaves categories blank, and tags looks an awful like categories so I adjusted it to scape the tags for this field.

Addresses issue #.
Fills in category when populating anidb table

-
Instead of parsing for categories from the AniDB API, since its non existent it parses tags for the categories


Bart39:
pull requests must be made against the dev branch and not master - please recreate your pull request against the correct branch

PR #2622, opened 2019-01-18
fix invalid updated datetime value

stanleykylee:
Addresses issue nZEDb#2621.

Per SO's question on datetime 0000-00-00-000000, it looks like 0000-00-00-000000 is actually invalid and causes the returned value of -0001-11-30 00:00:00. This fails inserts into the PreDB table.

Changing the initial updated datetime value to be the same as the created datetime value makes the most sense and prevents the invalid datetime value.


niel:
This doesn't fix the problem. The purpose of the update date is to see if the entry has been updated, i.e. if it is non-zero we can tell it has been updated at a glance, so setting it to the created date defeats the purpose of the filed.
Make sure your database has been set to allow ZERO_DATES, newer versions disable that by default.

PR #2638, opened 2019-09-19
Update NZBimport.php: fix group shortnames

ovpn-to:
sometimes nzbs contain shorted group names like: a.b.something and import script fails (script dies) without any error message because it tried to create the shortname group which often already exists as fullname and it dies right on "$groupID->save()"

steps to recall the bug:
create a NZB
add meta data group: a.b.something
add the group with fullname "alt.binaries.something" to nzedb
import the nzb from cli: "php NZBimport.php true true true 10"
import dies without an error or final message and no debugs shown.

patch needs more testing

### DIFF
--- a/nzedb/NZBImport.php
+++ b/nzedb/NZBImport.php
@@ -300,15 +300,18 @@ class NZBImport
 				$group = (string)$group;
 
 				// If groups_id is -1 try to get a groups_id.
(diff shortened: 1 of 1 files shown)


DariusIII:
Pull requests are accepted only to dev branch. Please create an PR for dev branch not 0.x.

PR #2640, opened 2019-09-19
Update NZBimport.php: fix group shortnames 

ovpn-to:
Update NZBimport.php: fix group shortnames

sometimes nzbs contain shorted group names like: a.b.something and import script fails (script dies) without any error message because it tried to create the shortname group which often already exists as fullname and it dies right on "$groupID->save()"

steps to recall the bug:
create a NZB
add meta data group: a.b.something
add the group with fullname "alt.binaries.something" to nzedb
import the nzb from cli: "php /var/www/nZEDb/misc/testing/nzb-import.php /path/ true true true 10"
import dies without an error or final message and no debugs shown.

### DIFF
--- a/nzedb/NZBImport.php
+++ b/nzedb/NZBImport.php
@@ -300,15 +300,18 @@ class NZBImport
 				$group = (string)$group;
 
 				// If groups_id is -1 try to get a groups_id.
+				// why ask if $groupID === -1 when we did not change it since it was set it in line 281?
 				if ($groupID === -1) {
+					// check here if groupname is a valid name, sometimes nzbs contains shorted groupnames like: a.b.something
+					$groupN = Groups::isValidName($group);
+					if($groupN !== false) $group = $groupN;
+					
 					if (array_key_exists($group, $this->allGroups)) {
 						$groupID = $this->allGroups[$group];
-						if (!$groupName) {
-							$groupName = $group;
-						}
+						if (!$groupName) $groupName = $group;
 					} else {
-						$group = Groups::isValidName($group);
 						if ($group !== false) {
+							$group = $groupN;
 							/*$groupID = $this->groups->add([
 								'name' => $group,
 								'description' => 'Added by NZBimport script.',
--- a/nzedb/NZBImport.php
+++ b/nzedb/NZBImport.php
@@ -317,6 +317,7 @@ class NZBImport
 						$groupID = $this->allGroups[$group];
 						if (!$groupName) $groupName = $group;
 					} else {
+						$group = Groups::isValidName($group);
 						if ($group !== false) {
 							/*$groupID = $this->groups->add([
 								'name' => $group,
--- a/nzedb/NZBImport.php
+++ b/nzedb/NZBImport.php
@@ -311,7 +311,6 @@ class NZBImport
 						if (!$groupName) $groupName = $group;
 					} else {
 						if ($group !== false) {
-							$group = $groupN;
 							/*$groupID = $this->groups->add([
 								'name' => $group,
 								'description' => 'Added by NZBimport script.',
--- a/nzedb/NZBImport.php
+++ b/nzedb/NZBImport.php
@@ -297,7 +297,7 @@ class NZBImport
 			// Get the group names, groups_id, check if it's blacklisted.
 			$groupArr = [];
 			foreach ($file->groups->group as $group) {
-				$group = (string)$group;
+				$group = strtolower($group);
 
 				// If groups_id is -1 try to get a groups_id.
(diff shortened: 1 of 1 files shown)


nzedb/NZBImport.php
```diff
@@ -297,15 +297,25 @@ protected function scanNZBFile(&$nzbXML, $useNzbName = false)
 			// Get the group names, groups_id, check if it's blacklisted.
 			$groupArr = [];
 			foreach ($file->groups->group as $group) {
-				$group = (string)$group;
+				$group = strtolower($group);
```
niel, line 300: We can't force to lowercase here. Uppercase letters are valid in group names.

nzedb/NZBImport.php
```diff
@@ -297,15 +297,25 @@ protected function scanNZBFile(&$nzbXML, $useNzbName = false)
...
 
 				// If groups_id is -1 try to get a groups_id.
+				// why ask if $groupID === -1 when we did not change it since it was set it in line 281?
 				if ($groupID === -1) {
```
niel, line 304: Probably because we're in a loop and the value may (should) be changed afterwards.
However it does feel wrong somehow, I'll look at it closer.

nzedb/NZBImport.php
```diff
@@ -297,15 +297,25 @@ protected function scanNZBFile(&$nzbXML, $useNzbName = false)
...
+                                        // test if group is a shortname
+                                        // $testgroup = "a.b.something";
+                                        if(preg_match("/^[a-zA-Z0-9]\.[a-zA-Z0-9]\./",$group)) {
```
niel, line 307: pregmatch is overkill, just do a sub-string test on the first four characters. a-zA-Z is redundant as you had used strtolower().

nzedb/NZBImport.php
```diff
@@ -297,15 +297,25 @@ protected function scanNZBFile(&$nzbXML, $useNzbName = false)
...
+                                                echo "group $group is a shortname\n";
```
niel, line 308: Remove this, I assume you added it for testing?

nzedb/NZBImport.php
```diff
@@ -297,15 +297,25 @@ protected function scanNZBFile(&$nzbXML, $useNzbName = false)
...
+                                                $groupN = Groups::isValidName($group);
+                                                if($groupN !== false){
+                                                        $group = (string)$groupN;
+                                                        echo "set new group, groupN = $groupN";
```
niel, line 312: Remove this too. We only report errors/unrecoverable problems in this method.

niel:
Also can you format the code according to our guidelines: http://li3.me/docs/specs/accepted/LSR-0-coding.md

nzedb/NZBImport.php
```diff
@@ -297,15 +297,25 @@ protected function scanNZBFile(&$nzbXML, $useNzbName = false)
...
+                                                        }
+                                        }
+					
 					if (array_key_exists($group, $this->allGroups)) {
 						$groupID = $this->allGroups[$group];
-						if (!$groupName) {
-							$groupName = $group;
-						}
+						if (!$groupName) $groupName = $group;
```
niel, line 318: Our coding style requires explicit braces, even for one-liners

nzedb/NZBImport.php
ovpn-to: ok but the system does only accept one type. 
you can't have group "a.b.something" in datebase when nzb contains a group named "a.b.SomeThing": import of this nzb fails. because it tries to create the group with capitals too which is not allowed. problem exists vice verse if you create group with capital letters like "test.MyGroup" for example and try to add another group "test.mygroup" you get a dead white page.

nzedb/NZBImport.php
ovpn-to: doh, maybe you're true, sometimes you don't see wood in the forest!

nzedb/NZBImport.php
ovpn-to: yes redundant since strtolower

nzedb/NZBImport.php
ovpn-to: still testing yes

nzedb/NZBImport.php
ovpn-to: only while testing

nzedb/NZBImport.php
ovpn-to: thanks will look into the link!

nzedb/NZBImport.php
niel: ok, I'll look into that.

PR #2643, opened 2019-10-23
Fix/charism/js lib errors

JReming85:
Addresses issue #.

When using Charisma there are many 404 errors for lib files.

- Fixed path to broken libs
- Added missing lib
- Fixed loading of woff2 fonts


Wally73:
it should be based on dev

PR #2644, opened 2019-10-23
Fix/charism/js lib errors

JReming85:
Addresses issue #.

When using Charisma there are many 404 errors for lib files.

Fixed path to broken libs
Added missing lib
Fixed loading of woff2 fonts


www/themes/Charisma/templates/basepage.tpl
```diff
@@ -38,7 +38,7 @@
 	<link href='{$smarty.const.WWW_THEMES}/{$theme}/css/elfinder.min.css' rel='stylesheet'>
 	<link href='{$smarty.const.WWW_THEMES}/{$theme}/css/elfinder.theme.css' rel='stylesheet'>
 	<link href='{$smarty.const.WWW_THEMES}/{$theme}/css/jquery.iphone.toggle.css' rel='stylesheet'>
-	<link href="{$smarty.const.WWW_THEMES}/shared/assets/pnotify/dist/pnotify.css" rel="stylesheet" type="text/css"/>
+	<link href="{$smarty.const.WWW_THEMES}/shared/libs/pnotify-3.0.x/dist/notify.css" rel="stylesheet" type="text/css"/>
```
niel: typo with the file name. Should be pnotify.css

JReming85: good catch, fixed

www/themes/Charisma/templates/basepage.tpl
```diff
@@ -210,6 +210,16 @@
 <!-- tinymce editor -->
 <script type="text/javascript"
 		src="{$smarty.const.WWW_THEMES}/shared/libs/tinymce-builded/js/tinymce/tinymce.min.js"></script>
+<!-- library for iphone style toggles -->
+<script type="text/javascript" 
+		src="{$smarty.const.WWW_THEMES}/{$theme}/js/jquery.iphone.toggle.js"></script>
+<!-- jQuery RATY Plugin -->
+<script type="text/javascript"
+                src="{$smarty.const.WWW_THEMES}/{$theme}/js/jquery.raty.js"></script>
```
niel, line 218: What is the purpose of these?

JReming85: When using the charisma theme, there are always 404/console errors for these missing .js files

Also 1 addition to allow woff2 fonts to be loaded as well otherwise .htaccess has it blocked.

pnotify example
image

Still looking for raty but it I think it may be related to rating and having stars visible.

My install is still a bit broken, but once I get it fully squared away I will attempt to fix/add/improve quite a bit

niel: Thanks for the clarification.

Would you prefer I wait for any additional fixes/improvements or proceed with the merge process?

JReming85: np, just reinstalled and waiting for pp to start working before I start adding my changes...

Here are the console errors I have noted with those these added
image

JReming85: I can go ahead and cancel this and reopen when I have all of charisma completed, that way there is no "incomplete" code/patches.

I mean its been this long without complaint so cant be that important, the only one I think is pressing is maybe the woff2 font in .htaccess but then again its been awhile with no complaint so I think it may be fine to wait

Issue #2653, opened 2020-04-14
xref column truncated data SQL Error


rbrostowiez:
###Expected behaviour
No errors in execution log of file: misc/update/nix/screen/threaded/start.sh

With a fresh install, and a single group enabled with backfill, I receive the below error message thousands of times per operation.

###Actual behaviour
SQL Error when inserting collections Error: SQLSTATE[22001]: String data, right truncated: 1406 Data too long for column 'xref' at row 1

###Steps to reproduce the behaviour
Execute: misc/update/nix/screen/threaded/start.sh

###Suggested solution:
At: `nzedb/Binaries.php:941` the xref column value being set is a concatenation of current value + up to 254 characters.  I have wrapped that concatenation with a substring call which will prevent column overflow to avoid the SQL Errors regarding the truncation of column data.

PR will be Submitted alongside this ticket.

PR #2654, opened 2020-04-14
Added a substring call to avoid the xref column truncate errors

rbrostowiez:
Addresses issue #2653.

- Wrapped xref column concatenation with a substring call to limit column value length.


ArroganTz (approved):

egyptianbman (approved):
Thanks!

